PR #76, opened 2014-11-27
Add ss protocol URI parse input

bygreencn:
Add ss protocol URI parse input.
Add regex rc and table crypty mode for parse_ss_uri.


clowwindy:
Implemented QRCode scan in e878b41

thanks

PR #117, opened 2015-01-07
Bring config form to front when it's opened before.

GangZhuo:
"configForm.Focus()" can't bring the form to front  when the form have been opened and back of some other windows. The solution is use "configForm.Activate()".


clowwindy:
Thanks!
Please remove the code you want to remove instead of just comment out it.

GangZhuo:
Removed. Please verify...

clowwindy:
Merged, thanks!

PR #118, opened 2015-01-07
Check update on independent thread

GangZhuo:
WebClient.DownloadStringAsync is very slow. This occurs because WebRequest by default detects and loads proxy settings the first time it starts, which can take quite a while. The solution is simply set the proxy property (WebRequest.Proxy) to null and it'll bypass the check. 

Set no proxy is bad way, because many companies through the proxy to the internet. So I use independent thread to check update.


clowwindy:
Why to start a new thread?

GangZhuo:
Maybe it only for me, it take me about 30 seconds. 
You can close it.
I use Windows 7 Ultimate SP1 64Bit.

PR #121, opened 2015-01-08
Logging timestamp

sinsinpub:
Simple implementation that adds a time stamp prefix to each log line.

### DIFF
--- a/shadowsocks-csharp/Controller/Local.cs
+++ b/shadowsocks-csharp/Controller/Local.cs
@@ -61,6 +61,7 @@ namespace Shadowsocks.Controller
         public void Stop()
         {
             _listener.Close();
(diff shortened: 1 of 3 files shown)


clowwindy:
Please also remove test code:

     // Console.WriteLine("Shadowsocks stopped");

sinsinpub:
It's not a 'test code', in fact, I added it after testing logging modifications.
I have commented it out 'cause I wonder if it is really necessary to log 'stop event'. Especially after adding the time-stamp, I think it should be closed like a 'start-stop tag' to make it clear: when the service stops. But just for restart event, it becomes annoying :disappointed:

clowwindy:
Thanks, merged

PR #122, opened 2015-01-08
Download new version automatically

kookxiang:
It will use proxy to check new version and download the latest version to temporary folder.
and it will open the zip file after download complete


clowwindy:
Currently we don't have any code signature or anything similar mechanics to protect against malware. Auto download and execution is dangerous. User interaction is required for now.

PR #187, opened 2015-03-10
Added support for assigning folder where ss_polipo.exe and polipo.conf are extracted

EraserKing:
This change aims to solve issues when system temp folder is not writable, or executing files from temp folder is blocked.

It added two command line arguments: /SameFolderForPolipo and /PolipoFolder.
If no arguments is used, ss_polipo.exe and polipo.conf are still extracted to temp folder as usual. User should not feel difference.
If /SameFolderForPolipo is used, these two files are extracted to the same folder where Shadowsocks.exe is located.
If /PolipoFolder <path> is used, these two files are extracted to the folder specified here.

### DIFF
--- a/shadowsocks-csharp/Controller/PolipoRunner.cs
+++ b/shadowsocks-csharp/Controller/PolipoRunner.cs
@@ -19,8 +19,33 @@ namespace Shadowsocks.Controller
 
         static PolipoRunner()
         {
-            // temppath = Path.GetTempPath();
-            temppath = Environment.CurrentDirectory;
+            temppath = Path.GetTempPath();
+            string[] commandLineArgs = Environment.GetCommandLineArgs();
+            for (int i = 0; i < commandLineArgs.Length; i++)
+            {
+                string commandLineArg = commandLineArgs[i];
+                if (commandLineArg.ToLower() == "/SameFolderForPolipo".ToLower())
+                {
+                    temppath = Environment.CurrentDirectory;
+                }
+                if (commandLineArg.ToLower() == "/PolipoFolder".ToLower())
+                {
+                    if (i == commandLineArgs.Length - 1)
+                    {
+                        System.Windows.Forms.MessageBox.Show("Polipo folder not specified.", "Error", System.Windows.Forms.MessageBoxButtons.OK, System.Windows.Forms.MessageBoxIcon.Error);
+                        throw new Exception("Polipo folder not specified");
+                    }
+                    else if (!IsDirectoryValid(commandLineArgs[i + 1]))
+                    {
+                        System.Windows.Forms.MessageBox.Show("Polipo folder not valid.", "Error", System.Windows.Forms.MessageBoxButtons.OK, System.Windows.Forms.MessageBoxIcon.Error);
+                        throw new Exception("Polipo folder not valid");
+                    }
+                    else
+                    {
+                        temppath = commandLineArgs[i + 1];
+                    }
+                }
+            }
             try
             {
                 FileManager.UncompressFile(temppath + "/ss_polipo.exe", Resources.polipo_exe);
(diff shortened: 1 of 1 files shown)


EraserKing:
Well, it seems AppVeyor ran into some trouble fetching source code...

shadowsocks-csharp/Controller/PolipoRunner.cs
```diff
@@ -20,6 +20,32 @@ class PolipoRunner
         static PolipoRunner()
         {
             temppath = Path.GetTempPath();
+            string[] commandLineArgs = Environment.GetCommandLineArgs();
+            for (int i = 0; i < commandLineArgs.Length; i++)
+            {
+                string commandLineArg = commandLineArgs[i];
+                if (commandLineArg.ToLower() == "/SameFolderForPolipo".ToLower())
```
wenhx: You should use String.Equals(string a, string b, StringComparison comparisonType) <https://msdn.microsoft.com/en-us/library/t4411bks(v=vs.110).aspx>

shadowsocks-csharp/Controller/PolipoRunner.cs
```diff
@@ -20,6 +20,32 @@ class PolipoRunner
...
             temppath = Path.GetTempPath();
+            string[] commandLineArgs = Environment.GetCommandLineArgs();
+            for (int i = 0; i < commandLineArgs.Length; i++)
+            {
+                string commandLineArg = commandLineArgs[i];
+                if (commandLineArg.ToLower() == "/SameFolderForPolipo".ToLower())
+                {
+                    temppath = Environment.CurrentDirectory;
+                }
+                if (commandLineArg.ToLower() == "/PolipoFolder".ToLower())
```
wenhx: Same as above.

shadowsocks-csharp/Controller/PolipoRunner.cs
```diff
@@ -20,6 +20,32 @@ class PolipoRunner
...
+                {
+                    if (i == commandLineArgs.Length - 1)
+                    {
+                        System.Windows.Forms.MessageBox.Show("Polipo folder not specified.", "Error", System.Windows.Forms.MessageBoxButtons.OK, System.Windows.Forms.MessageBoxIcon.Error);
+                        throw new Exception("Polipo folder not specified");
+                    }
+                    else if (!IsDirectoryValid(commandLineArgs[i + 1]))
+                    {
+                        System.Windows.Forms.MessageBox.Show("Polipo folder not valid.", "Error", System.Windows.Forms.MessageBoxButtons.OK, System.Windows.Forms.MessageBoxIcon.Error);
+                        throw new Exception("Polipo folder not valid");
```
wenhx: Is here and the line 36 thrown exception will crash the process?

shadowsocks-csharp/Controller/PolipoRunner.cs
```diff
@@ -122,5 +148,24 @@ private int GetFreePort()
             }
             throw new Exception("No free port found.");
         }
+
+        public static bool IsDirectoryValid(string path)
```
wenhx: This method does not appear to be called?

shadowsocks-csharp/Controller/PolipoRunner.cs
EraserKing: All right. I'll open a new pr later.

shadowsocks-csharp/Controller/PolipoRunner.cs
EraserKing: yes

shadowsocks-csharp/Controller/PolipoRunner.cs
EraserKing: see line 38

clowwindy:
I think this is not necessary since temp folder is not expected to be read only.

PR #197, opened 2015-03-20
Support Online PAC URL

GangZhuo:
see https://github.com/shadowsocks/shadowsocks-csharp/issues/163 .

One problem is that can't use blocked URL.


clowwindy:
First thanks for your work.

There're two problems:

1. You can just set the system PAC to the online PAC URL. No need to redirect.
1. You don't have implement a message box yourself. http://stackoverflow.com/questions/23693106/use-a-messagebox-to-get-a-user-input

GangZhuo:
new pull request at https://github.com/shadowsocks/shadowsocks-csharp/pull/198

clowwindy:
You don't have to use a new pull request. Just push new commits to this branch.

GangZhuo:
Ok, I will do it tonight.

GangZhuo:
Redirect maybe better, when share on LAN, others can get affect. Right?

clowwindy:
If users have specified an online URL, when they find the URL in their IE settings is still local URL, they will think it's a bug. And I don't think they want to use LAN sharing that way. If he already has an online PAC URL, he can just copy it to other devices.

Besides, some clients have compatibility issues on redirects. They if the PAC URL has an cache policy, they may also cache the 302 response, which is confusing to our users.

GangZhuo:
Done, please review.

shadowsocks-csharp/View/MenuViewController.cs
```diff
@@ -481,5 +491,38 @@ void openURLFromQRCode(object sender, FormClosedEventArgs e)
...
+        {
+            string origPacUrl = controller.GetConfiguration().pacUrl;
+            string pacUrl = Microsoft.VisualBasic.Interaction.InputBox(
+                I18N.GetString("Please input PAC Url"),
+                I18N.GetString("Update Online PAC URL"),
+                origPacUrl, -1, -1);
+            if (!string.IsNullOrEmpty(pacUrl) && pacUrl != origPacUrl)
+            {
+                controller.SavePACUrl(pacUrl);
+            }
```
clowwindy: You may need to warn the user if he didn't input a valid URL.

GangZhuo:
The InputBox also return empty string when user click the cancel button, so it is difficult to determine that the user input empty string or click the cancel button, maybe pop a confirm box?

clowwindy:
Well, if we can't distinguish empty string and cancel, I think we don't need validation. You can revert the last commit and I'll merge this pull request.

Thank you for the great work!

GangZhuo:
Done, please review.

clowwindy:
Merged, thanks!

GangZhuo:
Great, Thanks

PR #220, opened 2015-05-11
Add feature gfwlist2pac

everyx:
gfwlist2pac is a great tool and I think it could let shadowsocks-csharp more awesome.


everyx:
@clowwindy I found that this feature is contained in the previous version (https://github.com/shadowsocks/shadowsocks-csharp/commit/74c61ca76a3c032760c0b08190a6599b68ed4127), so maybe this feature is undesired for your, could you please tell me the resion? thanks!

clowwindy:
Thanks for the work, but we already support precise mode. Precise mode is good enough.

everyx:
I have some problerm of regexpRegExp like `"/^https?:\\/\\/[^\\/]+google\\.(.*)/"` in windows 8.1, maybe it is a bug?

PR #221, opened 2015-05-14
Add SocketAsyncEventArgs to handle local connections

xcnick:


clowwindy:
I think it's not necessary. It's just a client, not a busy server. It doesn't worth the effort to preallocate 8000 buffers of 4K size.

PR #223, opened 2015-06-07
规范多语言特性命名

celeron533:
Update the dictionary keys in languages file from nature English to consolidated names.


clowwindy:
I prefer the Apple localization style which uses the English version string directly as the key, making the code much easier to read and save binary space for an extra English version.

PR #256, opened 2015-08-06
Added list of servers in QRCode generate window to select.

JeffreyZhang105:
To add multiple servers in android phone, now it is no need to chose server one by one and generate QRCode for each. Just need to select the server listed in right.


clowwindy:
Merged, thanks

PR #279, opened 2015-08-17
Add gfwlist links

cicku:
Various fixes accompanied.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,30 +7,30 @@ Shadowsocks for Windows
 
 1. System proxy configuration
 2. PAC mode and global mode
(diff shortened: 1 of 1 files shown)


clowwindy:
Thanks, merged

clowwindy:
By the way, shall we update this URL?
https://github.com/shadowsocks/shadowsocks-windows/blob/master/shadowsocks-csharp/Controller/Service/GfwListUpdater.cs#L15

cicku:
I still push updates to google svn now, I will send a PR to you once it becomes read-only.

PR #300, opened 2015-08-19
improved `log viewer` windows

kimw:
improved `log viewer`:
1. add `clean log` feature
2. add `change font` feature
3. add `wrap text` feature
4. add `top most` feature
5. rename variables name to meaningful one. such as `menuItem2` -> `OpenLocationMenuItem`

### DIFF
--- a/shadowsocks-csharp/Controller/Service/UpdateChecker.cs
+++ b/shadowsocks-csharp/Controller/Service/UpdateChecker.cs
@@ -18,7 +18,7 @@ namespace Shadowsocks.Controller
         public string LatestVersionURL;
         public event EventHandler NewVersionFound;
 
-        public const string Version = "2.5.5";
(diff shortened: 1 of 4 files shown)


clowwindy:
Very useful features, thanks!

clowwindy:
By the way, what do you think if we turn the buttons and checkbox into menu items? We can add a `View` menu and put these functions into it.

kimw:
I think the toolbar is more efficient, but not most people need this efficiency. As you said, I've made some new menu items. And keep the old toolbar in hidden. Wish the people like it ;-)

BTW, I've been stuck in the scroll lock. I wanna add a feature to lock the log message while reading. And release the lock, or say, let it auto scrolling to the last log message when finished the reading. Any ideal?

See  [#306](/shadowsocks/shadowsocks-windows/pull/306)

PR #305, opened 2015-08-20
small change

Aivier:
1.Use MS YaHei as default font
2.Fix crash when add server after delete all server
3.When no server in the list,disable server info input box


kimw:
I'm adding the feature to save the default font and color and ... more into the config file. it'll be pull later.

clowwindy:
I have to warn you that I was invited for some tea yesterday. I won't be able to continue developing this project.

PR #310, opened 2015-09-05
Bugfix (maybe) & a new function

huhu2207:
Wrote a new function for Pinging servers before adding into the server
list, in case some guys input the invalid one.

BTW there is a bug in Configuration.cs. Why CheckServer function "check
itself"? I just commented it.


icylogic:
I don't think it's a good idea. We should not simply refuse to add a server according to a one-time ICMP test, since no server could ensure 0% package loss all day. In fact we are almost unable to check whether a server is available, and the ICMP test is the most unreliable method: What if the server to add is just down for a while? (Force user to wait for the server until it could response?) What if the server correctly response, but shadowsocks is actually unavailable (Incorrect iptables settings; wrong encryption method/password/port) ? What if the server intentionally disable Ping response?

If a user add a server that currently unusable, we could only leave detailed logs to help them find out the real problem.

What's more, the feature "availability statistics" added in 2.5.3 help user to find the most reliable server.

huhu2207:
Emm, you're right, I will return to previous edition and have a research on the availability statistics features

PR #312, opened 2015-09-07
Translation grammar fix & tiny bug fix

huhu2207:
Why the function CheckServer() runs in CheckServer() itself? Am I missed
looking up something?


GangZhuo:
It's overload method. https://msdn.microsoft.com/en-us/library/vstudio/ms229029(v=vs.100).aspx

PR #344, opened 2015-10-10
No privoxy mod

johnx7271:
Hi,
I am new to this project. Your guys are doing a great job on this project, and I appreciate it.
However, as this is a security sensitive product, so I think it would be great for this product to be self-contained. And I found this product depends on an external product: Privoxy, so I proceeded to try to remove the dependence on it. And this is my initial try. For more info, please see my modified readme.md.

I do not expect you to merge it with master. I just want to raise the issue and get in touch with you.

Besides the removal of Privoxy, I also added a little convenience: logging of net traffic flow among the client, the shadowshocks, and the final socks server. I think it may help developer checking the net flow and understanding what is really going through.

John Xie / Beijing


icylogic:
In fact, this product also depend on [SimpleJson](https://github.com/facebook-csharp-sdk/simple-json) from Facebook, zxing, and libsodium. We'll import more 3rd party libraries in upcoming updates. 
Since Privoxy is a open-source project which starts at 2001, I could not understand that you want to remove this because
>I found this product depends on an external product: Privoxy

As for net traffic logging, it's  a good idea. I suggest that you could separate these log from the main log for exceptions and errors. And if you like, you could open a standalone PR for this feature when you think it's stable and ready for use.

johnx7271:
Well, I know those dependences, they are of different type, one Is source code, another is executable. 
 
If someone in the contributors list has been involved with Privoxy project, that will give the public a better confidence of it. Are you aware of that?
 
The net traffic log is indeed separate from the main log, it can be turned on and off by a switch in the config file: logNetTraffic
 
John Xie
 
Date: Sat, 10 Oct 2015 01:22:15 -0700

icylogic:
I think you could maintain a non-privoxy version so that users will be able to make choice depend on whether they trust Privoxy.

johnx7271:
Yes, I intended to do that. But I am new to git and this project, nobody are following me, so I would like to have a new branch setup on origin, not on my own repository, so we can merge my work into that branch. My current work was not intended to merge with main, since it was originated from an earlier commit (2.5.8).
I do not have user right to do what I intended to do, so please help.

cs17899219:
### I vote to remove this executable type form main project. 

Although It is possible to run extra executable in main process, I still feel it is strange, the program must extra some executable file to your %temp% folder which means leave a lot of junk and run it. These actions are not safe enough because the sub-process "privoxy" will still running when the main process crashes.

Additionally, I can not be noticed before I read the release notes, when I run the program  I can find nothing to introduce HTTP proxy on menu. I event don't known which port the privoxy is listening.
In my opinion, The friendly way is introduce Privoxy serve as a third part plugin, gain the opportunity to  user of choose whether run it or not.

### New branch in future
I read this article because I am also looking for a solution to integrate HTTP proxy to shadowsocks for Max OS X. My preference goes to listen on new port and forward HTTP proxy to socks5  directly by using part of code from privoxy so that It can allow  HTTP proxy listen on separate port for each instance and completely portable.  I will turn to Windows Version as soon as I finish the MAX OS X version.

@johnx7271 , You are welcome for participate.

wongsyrone:
As mentioned above, Privoxy is [open-source](https://www.privoxy.org/sf-download-mirror/Sources/) project and the usage of it can be found in this codebase as well, for example, you can check https://github.com/shadowsocks/shadowsocks-windows/blob/master/shadowsocks-csharp/Controller/Service/PolipoRunner.cs, the file name might be misleading.

I'm closing this. Please feel free to open another PR.

PR #348, opened 2015-10-11
Optimize ip-api call

vladc:


icylogic:
Thanks. 

Collecting geolocation info is considered useless for most users and might bring risks, so it would be removed in the future.

PR #372, opened 2015-11-12
Update privoxy_conf.txt

ayanamist:
use `__POLIPO_BIND_PORT__` to avoid potential port conflict when another program is using 8123
`__POLIPO_BIND_PORT__` substitution is supported by [PolipoRunner.cs](https://github.com/shadowsocks/shadowsocks-windows/blob/d24548a04f4701dfe9f3e07d6305b5ba3b7a56f8/shadowsocks-csharp/Controller/Service/PolipoRunner.cs#L65)


GangZhuo:
Merged

PR #376, opened 2015-11-16
Fix error in open log file in debug mode.

10people:
not open log file in debug mode, cause error when open log.


chenshaoju:
:+1:  Looks good to me, but someone else must approve.

10people:
yep, i use this to check network stream in debug mode.

PR #413, opened 2016-01-04
a lot of commits

kimw:
It's a lot of commits since Aug, 2015.

--- updated 1/5/16
* SimpleJson => Newtonesoft.Json
* bandwidth counted


chenshaoju:
:+1: Looks good to me,but someone else must approve.

wongsyrone:
It will be better if commits get squashed.

GangZhuo:
@kimw  Can you update the code for: If a Log Form already opened, then active the Form instead of open a new Log Form?

kimw:
@GangZhuo It's a feature that the users/developers can view logs in seperated windows. Sometimes, it's usefull while the logs are rolling fast. On this feature, people can let a log rolling happylly in one logViwer window as it be (e.g. watching the newest messages), and stare on another non-rolling logViwer. It's very usefull at the time we trying to finger out network problems, that's what the logs main function I think so.

Nonetheless, if you insist on single-log-window feature, I can code it as an option. But, I keep my opinions.

Expecting your words.
---
Kim

GangZhuo:
@kimw Ok, you are right.

GangZhuo:
Merged. Thanks.

PR #447, opened 2016-02-20
Test connection speed

Mygod:
This PR is meant to show off a feature that basically works for me. :laughing:

What it does: for every config, switch to it, download a 1.53MB files in 4s or time out. Save the results to its `remark` and pick the fastest one.

Bugs/problems:
* No localization;
* It can create a huge number of threads if there are a lot of configs that doesn't work at all; (using `Thread.Abort` isn't a good practice, and it doesn't do much because most of the time the thread is stuck running unmanaged code)  
  This may lead to `OutOfMemoryException`. A possible resolution is to use `TaskFactory` to manage the threads and wait for them to finish when there are a large number of threads.
* No customization;  
  I only picked the config that works best for me.
* I use system console for output which is ugly, use a better UI;
* Terminating have some bug;
* No unit tests;
* Maybe more.

### DIFF
--- a/shadowsocks-csharp/Controller/ShadowsocksController.cs
+++ b/shadowsocks-csharp/Controller/ShadowsocksController.cs
@@ -252,7 +252,9 @@ namespace Shadowsocks.Controller
 
         public static string GetQRCode(Server server)
         {
-            string parts = server.method + ":" + server.password + "@" + server.server + ":" + server.server_port;
+            string parts = server.method;
+            if (server.auth) parts += "-auth";
+            parts += ":" + server.password + "@" + server.server + ":" + server.server_port;
             string base64 = Convert.ToBase64String(Encoding.UTF8.GetBytes(parts));
             return "ss://" + base64;
         }
--- a/shadowsocks-csharp/Model/Server.cs
+++ b/shadowsocks-csharp/Model/Server.cs
@@ -9,6 +9,12 @@ namespace Shadowsocks.Model
     [Serializable]
     public class Server
     {
+        public static readonly Regex
+            UrlFinder = new Regex("^ss://((?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?)$",
+                                  RegexOptions.Compiled | RegexOptions.IgnoreCase),
+            DetailsParser = new Regex("^((?<method>.+?)(?<auth>-auth)??:(?<password>.*)@(?<hostname>.+?)" +
+                                      ":(?<port>\\d+?))$", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+
         public string server;
         public int server_port;
         public string password;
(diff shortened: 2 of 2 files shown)


Mygod:
[Compiled x86 binary](https://mygod.tk/resources/userscloud/53geabxj0olf).

chenshaoju:
:-1: After this pull requests merged,The build is succeed,But is unusable,Shadowsocks.exe Crash at starting.Please review your code.

https://ci.appveyor.com/project/icylogic/shadowsocks-windows-l9mwe/build/1.0.108/job/he3p452nsnrusxat/artifacts

`问题签名:
  问题事件名称:	CLR20r3
  问题签名 01:	Shadowsocks.1.0.108.exe
  问题签名 02:	2.5.8.2
  问题签名 03:	56c82152
  问题签名 04:	Shadowsocks
  问题签名 05:	2.5.8.2
  问题签名 06:	56c82152
  问题签名 07:	460
  问题签名 08:	b
  问题签名 09:	System.IO.FileLoadException
  OS 版本:	6.1.7601.2.1.0.768.3
  区域设置 ID:	2052
  其他信息 1:	0a9e
  其他信息 2:	0a9e372d3b4ad19135b953a78882e789
  其他信息 3:	0a9e
  其他信息 4:	0a9e372d3b4ad19135b953a78882e789

联机阅读隐私声明:
  http://go.microsoft.com/fwlink/?linkid=104288&clcid=0x0804

如果无法获取联机隐私声明，请脱机阅读我们的隐私声明:
  C:\Windows\system32\zh-CN\erofflps.txt
`

Mygod:
I'm 100% sure running a single binary without dll dependencies will definitely crash (as it says FileLoadException). This is not this PR's fault. Download the compiled binary or compile it yourself.

chenshaoju:
My here is user side black box testing.
Can't push your build version to public.
All code must build on Appveyor ,and pass testing.
Just review your code please. :)

Licshee:
changes made by @Mygod was based on an earlier commit, which contains the bug @chenshaoju reported since then and eventually fixed in my fork

now the fix is merged into the master, please merge and try push again

Mygod:
Ok I will merge the changes later tonight. But beware that this PR still has several things needs fixing (which I'm too lazy to do) before merging and publishing a new version.

Licshee:
since the time "something happened" back then, "new versions" are really rare for now (if no completely eliminated)
for users really want new versions, they always download from AppVeyor

thus, it doesn't matter.

but still, merging is good for further developing, andthe owners read comments before taking action

so, no worries

Mygod:
Merged. Check it now.

Licshee:
yea, that's the point of using github
... and your commit works now, great

kimw:
Will this new feature cause some kind of disclose?

After all, the fixed download file (`https://dl-ssl.google.com/googletalk/googletalk-setup.exe`) has the fixed size of transfer.

Furthermore, it will cause network transfers through all you shadowsocks server of the same timespan.

Mygod:
You are right. I have never considered the security issues when implementing this. I did this for my very own convenience.

I mentioned "no customization" in one of my bugs/problems but now I think having a default set is probably a bad idea. This should probably never get merged but I'll leave this PR open for anyone to try it.

kimw:
@Mygod How about try to download a part of a big file with random size?

Licshee:
does speedtest.net have any open api can use?

Mygod:
@kimw The current implementation exits either the 1.53MB file has been downloaded or 4 seconds timeout expired. So it's kind of downloading a part of a file if your connection is slow.

chenshaoju:
You can not require download finish,example,When downloaded 5MB ,Then stop, And randomize the download size.Google chrome maybe is a good choose.

https://dl.google.com/update2/installers/ChromeStandaloneSetup.exe

Mygod:
As I've said, it's best to add customization to this feature. We can use 1MB to 2MB as default value but the user should be able to change this two values (and also the file they want to test). I'll now go implement the random file size now.

Mygod:
@Licshee I oppose using speedtest.net (at least as default) as it sometimes provides incorrect results and/or it doesn't reflect its performance on real world usage.

GangZhuo:
@Mygod  Can you reference blow code to implement features? 

* https://github.com/GangZhuo/shadowsocks-windows/blob/dev2/shadowsocks-csharp/Controller/Service/ServerTester.cs

* https://github.com/GangZhuo/shadowsocks-windows/blob/dev2/test/ServerTesterHelper.cs

* https://github.com/GangZhuo/shadowsocks-windows/blob/dev2/test/UnitTest.cs#L196

Mygod:
@GangZhuo Did you mean I should fork your repo, make the changes and create a PR there instead?

GangZhuo:
@Mygod  No, just copy my code, and then make your own changes. Or, just reference.

PR #537, opened 2016-04-15
Connection test prototype

Mygod:
Need to use master branch. Closes #447.


chenshaoju:
Appveyor build failed.Please review your code:
https://ci.appveyor.com/project/icylogic/shadowsocks-windows-l9mwe/build/1.0.136/job/39e2e13lp05gajv9

Mygod:
The remote name could not be resolved: 'www.nuget.org'

Wow thanks for blaming me for that. I'm 100% happy you did that.

chenshaoju:
@Mygod My fault,I think I probably need quit alcohol 😅

wongsyrone:
It reports "dead" against a working server, can you check it?

Mygod:
Maybe my standards are a little bit too aggressive. You can retry that or tweak the settings.

Mygod:
@chenshaoju Take [this useful utility](https://github.com/noidontdig/gitdown). It's for your own good. :-)

Mygod:
I have updated this demo to work with latest master commits. Feel free to reopen it or keep this closed.

chenshaoju:
Compile is successful.

But looks like a prototype, No GUI and when testing complete, The console window immediately disappear...

Mygod:
Yeah it's more like a script that I haven't used for two years (and probably won't be using for another two years). I have no intention of doing GUIs.

Test results are saved to remarks (overwritten). It's a quick hack.

celeron533:
Thanks @Mygod 
Could also use it as a connectivity test with a smaller size request as connectivity diagnosis. If you do not have time to do the UI part, I'll handle it.

Mygod:
In [my original commit](https://github.com/shadowsocks/shadowsocks-windows/pull/537/commits/df2a0d95d9ee9403e2604d11cfa47ea39a2d5401), I used [googletalk-setup.exe](https://dl-ssl.google.com/googletalk/googletalk-setup.exe) which is 1.53MB. I mainly use this tool to test streaming speed.

As mentioned [here](https://github.com/shadowsocks/shadowsocks-windows/issues/1647#issuecomment-361038999), if you want to merge this, I'd suggest making test URL and timeouts configurable to minimize risks. And I'd suggest not making this periodic.

celeron533:
@Mygod Yes I am also worried that periodic behavior. Any fixed size, fixed query and repeatedly actions should be avoided.

Mygod:
A somewhat quick (but should work fine) workaround for fixed size queries is to fetch a URL with random length like appending `?<string of random length>`, and choose a file that's too big to finish under normal conditions within say 4-6 seconds, and cut off the download after that period of time.

celeron533:
PS. Check for updates on startup should already be a kind of fingerprint...

database64128:
If no one intends to maintain this feature, we are removing the related GUI from v4 and dropping it completely in v5.

PR #538, opened 2016-04-15
Support one-time authentication in ss urls for sharing

Mygod:
This commit implements the new ss protocol proposed here:
https://github.com/shadowsocks/shadowsocks-org/issues/16

Also removes a TODO in Server.cs. You are welcome.


GangZhuo:
Thanks

PR #551, opened 2016-05-01
Disable StatisticsStrategy by default.

kimw:
As the StatisticsStrategy is not stable, it should be disable by default.


chenshaoju:

kimw:
bump 3.1 to force upgrade with *disable StatisticsStrategy by default*

GangZhuo:
Thanks

PR #555, opened 2016-05-03
fixed bandwidth counting by servers

kimw:
The bandwidth counting previously I PRed has bug that it can't record bandwidth by each server. Now it's fixed.

The bandwidth is record in config file like follows,
```
{
  "configs": [
    {
      "server": "1.2.3.4",
      "remarks": "My SS Server #1",
      ...
      "bandwidthIn": 46336870,
      "bandwidthOut": 904840
    },
    {
      "server": "a.b.c.d",
...
```

### DIFF
--- a/shadowsocks-csharp/Controller/Service/Listener.cs
+++ b/shadowsocks-csharp/Controller/Service/Listener.cs
@@ -78,7 +78,7 @@ namespace Shadowsocks.Controller
                 _tcpSocket.Listen(1024);
 
                 // Start an asynchronous socket to listen for connections.
-                Logging.Info("Shadowsocks started");
+                Logging.Info($"Shadowsocks started, using [{_config.GetCurrentServer().FriendlyName()}].");
                 _tcpSocket.BeginAccept(new AsyncCallback(AcceptCallback), _tcpSocket);
                 UDPState udpState = new UDPState();
                 _udpSocket.BeginReceiveFrom(udpState.buffer, 0, udpState.buffer.Length, 0, ref udpState.remoteEndPoint, new AsyncCallback(RecvFromCallback), udpState);
--- a/shadowsocks-csharp/Controller/Service/TCPRelay.cs
+++ b/shadowsocks-csharp/Controller/Service/TCPRelay.cs
@@ -69,14 +69,14 @@ namespace Shadowsocks.Controller
             return true;
         }
 
-        public void UpdateInboundCounter(Server server, long n)
+        public void IncreaseInboundCounter(Server server, long n)
         {
-            _controller.UpdateInboundCounter(server, n);
+            _controller.IncreaseInboundCounter(server, n);
         }
 
-        public void UpdateOutboundCounter(Server server, long n)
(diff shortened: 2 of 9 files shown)


GangZhuo:
There are some issues:
* Can't restore the count record, since SS do not save the record to the config file when exit.
* The count record will be lose, when user edit server(s) by Config Form or close Log View Form. 

Sorry for bad english.

kimw:
@GangZhuo 
Thanks for your reply. I'll check the issues ASAP.
BTW, I can read in Chinese.

kimw:
@GangZhuo 
As you replied,
> Can't restore the count record, since SS do not save the record to the config file when exit.

How can I re-happen it?

GangZhuo:
@kimw Steps:

1) Open gui-config.json file, and remember the values of bandwidthIn and bandwidthOut.

2) Start the shadowsocks, and do nothing (Do not open the Log Form).

3) Open a website which is blocked by gfw.

4) Right-click the taskbar icon and select Quit to quit the shadowsocks .

5) Review the gui-config.json file, the values of  bandwidthOut and bandwidthIn should be changed, but there are not.

kimw:
@GangZhuo The bugs that you reported is fixed. Thanks for your help.

>There are some issues:
> - Can't restore the count record, since SS do not save the record to the config file when exit.

fixed in https://github.com/shadowsocks/shadowsocks-windows/pull/555/commits/356a3a1fb0ec0fcbbd42a685a92f61fb12339eca

> - The count record will be lose, when user edit server(s) by Config Form or close Log View Form.

fixed in https://github.com/shadowsocks/shadowsocks-windows/pull/555/commits/0ca4aa8f0370d39074511a98ea3ba5cda3e48606

Mygod:
I prefer to put the traffic statistics in the statistics config file. What do you think?

kimw:
@Mygod, I did not use statistics module. So I'm not sure that is statistics module stable?

Mygod:
I feel like bandwidth statistics doesn't fit in config.json. One should put only configurations in that file.

kimw:
@Mygod

It's a big different design philosophy between a *single big config file* and a *series of small config files*. Single-config-file is more simple to manage but messed; series-config-files are more understandable but not easy to manage. People like me sometimes, maybe almost the time, ignore the non-main config file by no reasons unless someone notice me. Let's suppose that, some day in the future, the `shadowsocks` has 10 features. Woo, there will be 10 config files separately. I must crazy!

On the other side, separated config files are indeed clear to understand. One config file has always one meaning. But I have to config different functions in different files.

Til now, my personal opinion is that, the `config.json` is a JSON formatted file which can be enlarge/shrink freely. In that means, single-config-file is much flexible at the time now.

Limited by my knowledge, I don't have the final result on this topic. But your question makes me think.

What do you think about? Can you share it?

Mygod:
Well the point is that statistics are NOT config. It's generated by the program instead of input from user.

wongsyrone:
If you want to get this merged, please follow advice talked above and solve conflicts.

wongsyrone:
Close due to inactive.

PR #566, opened 2016-05-15
New Feature: print "connect to hostname:port" to log

kimw:
- featured print "connect to hostname:port" to log
- renamed private properties with prefix underline
- clean up code in C# 6
- clean up buggy parameter.
  - Replaced SocketFlags.None (an enum which euqals 0) from zero.

### DIFF
--- a/shadowsocks-csharp/Controller/Service/TCPRelay.cs
+++ b/shadowsocks-csharp/Controller/Service/TCPRelay.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Net;
 using System.Net.Sockets;
 using System.Timers;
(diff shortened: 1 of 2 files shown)


shadowsocks-csharp/Controller/Service/TCPRelay.cs
```diff
@@ -586,39 +489,52 @@ private void PipeRemoteReceiveCallback(IAsyncResult ar)
...
                 {
+                    int atyp = _connetionRecvBuffer[0];
+                    string dst_addr;
+                    int dst_port;
+                    switch (atyp)
+                    {
+                        case 1:  // IPv4 address, 4 bytes
+                            dst_addr = new IPAddress(_connetionRecvBuffer.Skip(1).Take(4).ToArray()).ToString();
+                            dst_port = (_connetionRecvBuffer[5] << 8) + _connetionRecvBuffer[6];
+                            Logging.Debug($"connect to {dst_addr}:{dst_port}");
```
wongsyrone: Only Debug version will print out these, Release version will strip out them.
ref: https://github.com/shadowsocks/shadowsocks-windows/blob/master/shadowsocks-csharp/Controller/Logging.cs#L51

wongsyrone:
Other minor changes locate in https://github.com/wongsyrone/shadowsocks-windows/commits/kimw-log_connect_to, if no objections, I will merge this into master.

PR #573, opened 2016-05-18
Fixed the PAC to make it work with Firefox x64

### LINKED ISSUE #487: 请问PAC失效的问题在3.0版本中是否已解决了？
请问PAC失效的问题在3.0版本中是否已解决了？ 谢谢！

kimw:
fixed #487 #439 #367 #270 

use the new `Fast Mode` to fix the issues above.


everyx:
之前的版本是支持这个的，后来去除了该功能，现在支持自定义 pac 模板，这个可能没有太大必要

kimw:
Hi @everyx,

1. I've look up the commit log, but haven't found the one that _committed before but removed later_. Can you point it out?
2. Do you use Firefox x64 on Windows? In which, the PAC file itself should be sized tiny to make it work.

everyx:
See #220

kimw:
@everyx Sorry for delay.

First, yes, it's definitely a FF's (all the 64-bit versions) bug. My point is that we can adjust SS code till the bug is fixed in FF, or else the PAC function will not work. *MAKE IT WORK* is the most important thing, I think, if I stand on the user side.

And, as replied in #220, you and @clowwindy were talking about is it suitable to import fast mode into this repo. Is it a different topic from this one? In fact, at the time, May 12, 2015, the FF is work well with PAC both precise mode and fast mode. Besides the precise mode can pattern on full URL, but fast mode patterns on domain only. If FF can work with PAC right now, I must got the same idea to choose not to import fast mode as you.

Finally, you must every good at the source code. WTF, there's a guy who can even pick out the deleted code from commit history. Thumb you up! I have to!

chenshaoju:
This patch has tested on Windows 10 + Firefox 46.0.1,And it worked.
:+1: Looks good to me, but someone else must approve.

xeonchen:
@kimw Would you give me more information about the bug of Firefox if it still exists?

wongsyrone:
Please test latest Firefox, this issue should be fixed in FF instead of SS in my opinion.
BTW, this PR has conflicts to be resolved.

chenshaoju:
2016-08-08 Test result:
Using Shadowsocks 3.0 release version.

Firefox 48.0 ,PAC is not work:



Firefox 51.0a1 (Nightly version) ,PAC is work well:
51 0a1

wongsyrone:
Thanks for testing. I'm closing this.

PR #604, opened 2016-06-29
SS over proxy support

Noisyfox:
At my working place, I need to use a socks5 proxy first in order to connect to the internet.
So I made this mod to add proxy support. Now ss can connect to the server through a socks5 proxy.
It's possible to add more proxy support such as http proxy and socks4.
And also you can use ss over another ss.
It works pretty fine on my computer but it may need more test.


wongsyrone:
@Noisyfox Can you add something to choose proxy type? User might get confused after opening the ProxyForm.

wongsyrone:
If you want to get this merged, please solve conflicts

Noisyfox:
@wongsyrone updated

wongsyrone:
Can you replace all "Logging.Debug()" to something like:
```cs
if (config.isVerboseLogging) {
    Logging.Info();
}
```
to make use of newly added "verbose logging" feature?
ref: https://github.com/shadowsocks/shadowsocks-windows/commit/fb3fcdb90d68ad6e8c582ecb5d70c26103017fe4 and https://github.com/shadowsocks/shadowsocks-windows/commit/d3acb3e681b12085e5a4399547aee074853aa66f

Noisyfox:
You mean all those debug info related to the connection status?

wongsyrone:
Since Logging.Debug() is under conditional compiling, I think it is a good idea to let them debug without compiling Debug version by hand. Someone even doesn't know how to compile and debug.

Noisyfox:
So that's literally ALL Debug calls? Among the whole project or just in TCPRelay?

wongsyrone:
You decide which one should be verbose, which should belong to programmer.

Noisyfox:
Well... It's quite verbose now lol

PR #641, opened 2016-08-05
Fix the System.ObjectDisposedException when closing the LogForm

celeron533:
This exception is caused by processing Invoke Action while trafficChart is already disposed.

=====
Exception reproduce frequency: RARE. Once triggered, ShadowSocks crashes.


wongsyrone:
Confirmed by issue reporter, merged.
Thanks for quick response.

PR #660, opened 2016-08-15
Resolving domain name through socks5 proxy

Noisyfox:
When using ss through another socks5 proxy, the prev version still resolves ss server's domain name locally. This could lead to wrong results.

This patch allow proxy module to resolve domain name itself.
-Socks5 proxy: just pass the domain name to socks5 proxy server cuz the protocol support domain name.
-Direct connect: still resolves locally and synchronously.

### DIFF
--- a/shadowsocks-csharp/Controller/Service/TCPRelay.cs
+++ b/shadowsocks-csharp/Controller/Service/TCPRelay.cs
@@ -302,8 +302,7 @@ namespace Shadowsocks.Controller
                 if (ar.AsyncState != null)
                 {
                     connection.EndSend(ar);
-                    // TODO: need fix
-                    //Logging.Debug(remote.LocalEndPoint, remote.DestEndPoint, RecvSize, "TCP Relay");
+                    Logging.Debug(remote.LocalEndPoint, remote.DestEndPoint, RecvSize, "TCP Relay");
                     connection.BeginReceive(_connetionRecvBuffer, 0, RecvSize, SocketFlags.None, new AsyncCallback(ReadAll), null);
                 }
                 else
@@ -311,8 +310,7 @@ namespace Shadowsocks.Controller
                     int bytesRead = connection.EndReceive(ar);
                     if (bytesRead > 0)
                     {
-                        // TODO: need fix
-                        //Logging.Debug(remote.LocalEndPoint, remote.DestEndPoint, RecvSize, "TCP Relay");
+                        Logging.Debug(remote.LocalEndPoint, remote.DestEndPoint, RecvSize, "TCP Relay");
                         connection.BeginReceive(_connetionRecvBuffer, 0, RecvSize, SocketFlags.None, new AsyncCallback(ReadAll), null);
                     }
                     else
@@ -344,8 +342,7 @@ namespace Shadowsocks.Controller
         private class ProxyTimer : Timer
         {
             public IProxy Proxy;
-            public string DestHost;
-            public int DestPort;
+            public EndPoint DestEndPoint;
             public Server Server;
 
             public ProxyTimer(int p) : base(p)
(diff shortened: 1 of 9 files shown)


shadowsocks-csharp/Proxy/Socks5Proxy.cs
```diff
@@ -73,48 +74,62 @@ public void EndConnectProxy(IAsyncResult asyncResult)
...
+                    case AddressFamily.InterNetwork:
+                        request = new byte[4 + 4 + 2];
+                        atyp = 1; // IP V4 address
+                        break;
+                    case AddressFamily.InterNetworkV6:
+                        request = new byte[4 + 16 + 2];
+                        atyp = 4; // IP V6 address
+                        break;
+                    default:
+                        throw new Exception(I18N.GetString("Proxy request faild"));
```
wongsyrone: typo, and you need to add this string to Data/cn.txt

shadowsocks-csharp/Controller/Service/TCPRelay.cs
```diff
@@ -302,15 +302,17 @@ private void ReadAll(IAsyncResult ar)
                 if (ar.AsyncState != null)
                 {
                     connection.EndSend(ar);
-                    Logging.Debug(remote.LocalEndPoint, remote.DestEndPoint, RecvSize, "TCP Relay");
+                    // TODO: need fix
```
wongsyrone: Do you mean "don't merge this until fixed"?

Noisyfox:
Fixed

wongsyrone:
Can we use something like:
```C#
if (typeof(Socket).GetProperty("DualMode") == null)
```
to detect DualMode availibility, and prefer to use it?

Noisyfox:
Well I think the main problem is the version of windows not the version of .net api.
According to [this article](https://msdn.microsoft.com/en-us/library/windows/desktop/bb513665(v=vs.85).aspx)

> In order to support both IPv4 and IPv6 on Windows XP with Service Pack 1 (SP1) and on Windows Server 2003, an application has to create two sockets, one socket for use with IPv4 and one socket for use with IPv6. These two sockets must be handled separately by the application.
>Windows Vista and later offer the ability to create a single IPv6 socket which can handle both IPv6 and IPv4 traffic.

If running on windows xp, we still need to handle ipv4 and ipv6 separately. And if running on vista or later os, even if old .net version doesn't have a "DualMode" property, we could still use `socket.SetSocketOption(SocketOptionLevel.IPv6, (SocketOptionName)27, false);` to enable dual-mode.

wongsyrone:
Dropping WinXP support seems reasonable to me. Most of them are using Win7.

Noisyfox:
Maybe we could also update to .net 4.5 :)

Noisyfox:
By the way on the [official download page](https://shadowsocks.org/en/download/clients.html) it says "Windows 7 or above" :)

wongsyrone:
1. Use OperatingSystem.Version to detect version, If it is "5.1.x" then exit with error message indicating to upgrade system.
2. It will use .NET4 or 4.5 if available  
ref: https://msdn.microsoft.com/en-us/library/jj152935(v=vs.110).aspx and https://msdn.microsoft.com/en-us/library/w4atty68(v=vs.110).aspx

Noisyfox:
Yeah the version of .net 4 or 4.5 is not a big problem.

Noisyfox:
I see 
`  <supportedRuntime version="v4.0" />
  <supportedRuntime version="v2.0.50727" />`
in app.config. That means we also support .net 3.5?

wongsyrone:
It will prefer to use .NET4 or above. If not found, then use 3.5.

Noisyfox:
So it may not a good idea to use those features don't exist in .net 3.5?

wongsyrone:
Drop old version string in app.config or detect .NET version on runtime. I'm not sure which one is better.

Noisyfox:
Well... Since a clean windows 7 only support .net 3.5 and below, maybe it's not a good idea to drop this version.

wongsyrone:
Found an official way to get .NET version: https://msdn.microsoft.com/en-us/library/hh925568(v=vs.110).aspx

wongsyrone:
Added some checks to ensure all functionalities working well, if no objections, I will merge this.
https://github.com/wongsyrone/shadowsocks-windows/commits/Noisyfox-dev_resolving

wongsyrone:
Merged, thanks.

Issue #679, opened 2016-08-24
新版本（3.2）崩溃


sunday1103:
新版本经常崩溃，这是最近一次弹出的信息。（以前没有弹出过）

![image](https://cloud.githubusercontent.com/assets/3799028/17923133/308444b8-6a15-11e6-91fa-d2386d9874ff.png)

系统状况：
![image](https://cloud.githubusercontent.com/assets/3799028/17923176/6e49a59a-6a15-11e6-8d2c-99614fd31590.png)

chenshaoju:
请尝试该版本：
https://ci.appveyor.com/project/icylogic/shadowsocks-windows-l9mwe/build/job/uvu9en8tpxn5c2pe/artifacts

wongsyrone:
Please try latest Appveyor version.

Issue #664, opened 2016-08-17
刚刚出现一个运行时错误


772807886:
![default](https://cloud.githubusercontent.com/assets/10294977/17736240/3d62c6e8-64b9-11e6-8ea3-5634731863c8.PNG)

772807886:
便携模式，Win 10，14393

772807886:
版本3.2

wongsyrone:
> although all methods of ConcurrentDictionary<TKey, TValue> are thread-safe, not all methods are atomic, specifically GetOrAdd and AddOrUpdate. 

This might be the root cause.

wongsyrone:
Please test this version: https://ci.appveyor.com/project/wongsyrone/shadowsocks-windows/build/1.0.32/job/nd6t47crxlglbu97/artifacts

codexss:
我也遇到了
![image](https://cloud.githubusercontent.com/assets/5959201/17765722/2d95965e-655a-11e6-9005-34d71c25fc88.png)

wongsyrone:
How about the link I posted above?

codexss:
@wongsyrone 看起来好像是开启负载均衡之后出现的问题……

wongsyrone:
I'm not sure whether I really fix the issue, please test version I posted above.

codexss:
@wongsyrone 上面那个版本也有这个问题

wongsyrone:
What's your .NET version?

codexss:
.net 4.0 win10 x64

Syrone Wong <notifications@github.com>于2016年8月18日周四 19:36写道：

Noisyfox:
I think it may be caused by the fact that "_lastInboundCounter" and "_lastOutboundCounter" are not guaranteed to be updated in pairs (They are updated separately at [here](https://github.com/shadowsocks/shadowsocks-windows/blob/a1eef6eac1499ad9baf985bb5d6fa3a6c2b4246c/shadowsocks-csharp/Controller/Service/TCPRelay.cs#L556) and [here](https://github.com/shadowsocks/shadowsocks-windows/blob/a1eef6eac1499ad9baf985bb5d6fa3a6c2b4246c/shadowsocks-csharp/Controller/Service/TCPRelay.cs#L628)). 
[...]

PR #682, opened 2016-08-26
New approach to record and calculate inbound/outbound speed.

Noisyfox:
Use lock to sync between threads.This should fix issue #679 
and other same issue.

Fix a bug in UpdateSpeed() that only save the first xxxSpeedRecord.
This bug was introduced in commit 2aa0620


Noisyfox:
#664

chenshaoju:
Quick test is pass,Base appveyor version.

wongsyrone:
LGTM.

PR #687, opened 2016-08-29
Fix #561

### LINKED ISSUE #561: Empty chunk when OTA is enabled.
If OTA is enabled, the first chunk may be empty. It just contain data length 0 and the Hmac-SHA1 result.

Noisyfox:
With some refactor.


wongsyrone:
Merged into master branch.

PR #696, opened 2016-09-05
Add KCP support

chrislonng:
Add KCP support for Shadowsocks.
KCPTUN should be setup on a different port to avoid conflicts with Shadowsocks' UDP support.

refs: https://github.com/xtaci/kcptun
refs: https://github.com/shadowsocks/shadowsocks-android/pull/788


wongsyrone:
To be honest, I'm not willing to merge this. Shadowsocks itself should keep simple; it is not a good idea to involve other irrelevant programs into the codebase. If the user wants this, they can choose their own way. The last reason is the release cycle of kcptun; it is too frequent in my opinion, sometimes the new one does not compatible with the older one. We have to keep up with it all the time. It is a life killer. For Android version, please refer to https://github.com/shadowsocks/shadowsocks-windows/issues/644#issuecomment-238067370

chrislonng:
Agree with you, Shadowsocks should keep simple. +1 :smile: 
Someone who needs this feature could pull from my repo.

wongsyrone:
Other kcptun configuration GUI can be found at https://github.com/GangZhuo/kcptun-gui-windows and https://github.com/xtaci/kcptun/issues/161.

Thank you anyway. Closing.

PR #703, opened 2016-09-07
new traffic icon style

breakwa11:


wongsyrone:
LGTM. Any opinion on this?

Noisyfox:
Great.

wongsyrone:
@celeron533 What about yours?

shadowsocks-csharp/View/LogForm.Designer.cs
```diff
@@ -268,11 +268,13 @@ private void InitializeComponent()
             this.trafficChart.Palette = System.Windows.Forms.DataVisualization.Charting.ChartColorPalette.None;
             series1.ChartArea = "ChartArea1";
             series1.ChartType = System.Windows.Forms.DataVisualization.Charting.SeriesChartType.Line;
+            series1.Color = System.Drawing.Color.FromArgb(((int)(((byte)(255)))), ((int)(((byte)(128)))), ((int)(((byte)(0)))));
```
wongsyrone: It seems the type casting is redundant.
ref: https://msdn.microsoft.com/en-us/library/cce5h557(v=vs.110).aspx

celeron533: Agree.

PR #708, opened 2016-09-10
Add Hotkey Support

772807886:
#706 
Add hotkey support, there is a menu item in shadowsocks' menu to set up hotkeys.
Now it can switch enable status, mode, LAN Listener, switch server, and open the log window.


wongsyrone:
Please write commit title and description in English.

wongsyrone:
Also, write comments in English

shadowsocks-csharp/View/HotkeySetting.cs
```diff
@@ -0,0 +1,126 @@
...
+                if(e.Shift) {
+                    sb.Append("Shift + ");
+                }
+                if((e.KeyValue >= 33 && e.KeyValue <= 40) ||
+                    (e.KeyValue >= 65 && e.KeyValue <= 90) ||
+                    (e.KeyValue >= 112 && e.KeyValue <= 123)) {
+                    sb.Append(e.KeyCode);
+                } else if(e.KeyValue >= 48 && e.KeyValue <= 57) {
+                    sb.Append('D').Append((char)e.KeyValue);
+                } else if(e.KeyValue >= 96 && e.KeyValue <= 105) {
```
breakwa11: what these numbers means?

772807886: It mean the key code.
33~40 is the function keys on keyboard.
65~90 means A to Z on keyboard.
112~123 means F1 to F12 on keyboard.
48~57 means 0 to 9 number keys above letter keys on keyboard.
96~105 means 0 to 9 number keys at right of the keyboard.
You can google "Keycode" to get more information.

772807886:
I rebased the commits with commit title and description in English.

shadowsocks-csharp/View/HotkeySetting.cs
breakwa11: I means you should use constant instead

772807886: OK, I will change it.

shadowsocks-csharp/Controller/HotKey.cs
```diff
@@ -0,0 +1,116 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Windows.Forms;
+
+namespace Shadowsocks.Controller {
+    public class HotKey : Form {
+        /// <summary>
+        /// 接受热键的窗体
```
wongsyrone: These code comments should also in English.

wongsyrone:
I think about your implementation, finally, come up with another idea.

We don't have to handle Win API ourselves, https://github.com/kirmir/GlobalHotKey is handy after reading his code, it is in NuGet packages. Shadowsocks is a tray application; it is hard to maintain Form's handle, although creating a hidden form is possible.

You can rewrite your code based on https://github.com/wongsyrone/shadowsocks-windows/commits/shortcut, or in other ways follow the idea. I want this feature behaves like QQ's setting hotkey form.

Any idea?

772807886:
Ok, I merged your branch into my pull request.

wongsyrone:
I pushed a commit to finalize hotkey format, you can check commit message and the test cases.

wongsyrone:
Did you delete all your changes when rebasing?

772807886:
I had backup the folder before I rebase.
After rebase, all of changes was deleted.

wongsyrone:
Close due to conflicts and inactive.
You can check https://github.com/shadowsocks/shadowsocks-windows/pull/724.

I changed the server switching logic. Yours won't work if the user configured more than ten servers. I change this to move to prev/next server. This is a workaround to satisfy users who add many servers for backup.

Also, a user should register all hotkeys explicitly when starting Shadowsocks, since environment may vary. If you want to open multiple instances of Shadowsocks, you have to change hotkeys for another instance. Otherwise, it will not work.

PR #717, opened 2016-09-14
Refine Encryption module

breakwa11:
- `EncryptorInfo` struct instead of `Dictionary<string, int[]>>`
- <del>Remove useless debug info (also it cost too much time to refresh)</del>


wongsyrone:
Debug info will be stripped out from Release version; I don't think it makes sense.
Ref: https://github.com/shadowsocks/shadowsocks-windows/blob/master/shadowsocks-csharp/Controller/Logging.cs#L66

PR #721, opened 2016-09-15
Avoid NullReferenceException

breakwa11:


wongsyrone:
proxy config fixed via https://github.com/shadowsocks/shadowsocks-windows/commit/c1ec851627b9bba346896b1984c529ca9cb7db7a

breakwa11:
c1ec851 incorrect. U should test with an old version of user-config.json

wongsyrone:
Done in https://github.com/shadowsocks/shadowsocks-windows/commit/33b2ca8c8a44d0615f02b21cbd882b039f1f7426

PR #724, opened 2016-09-17
Add hotkeys

wongsyrone:


wongsyrone:
Usage:

How to input keys:
1. put focus in the corresponding textbox
2. press the key combination you want to use
3. when you think it is ready, release all keys
4. the textbox shows your input

How to change keys:
1. put focus in the corresponding textbox
2. press BackSpace to clear content
3. re-input new one

How to deactivate:
1. clear content in the textbox
2. press OK button

Meaning of label color:
- Green: this combination is not occupied by other programs and register successfully
- Yellow: this combination is occupied by other programs and you have to change to another one
- Transparent without color: initial status

PR #745, opened 2016-09-22
Add HTTP proxy support in forward proxy

Noisyfox:


shadowsocks-csharp/Proxy/HttpProxy.cs
```diff
@@ -0,0 +1,199 @@
...
+        }
+
+        public void EndConnectProxy(IAsyncResult asyncResult)
+        {
+            _remote.EndConnect(asyncResult);
+            _remote.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.NoDelay, true);
+        }
+
+        private const string HTTP_CRLF = "\r\n";
+        private const string HTTP_CONNECT_TEMPLETE =
```
wongsyrone: typo: TEMPLATE

shadowsocks-csharp/Controller/Service/TCPRelay.cs
```diff
@@ -420,7 +420,18 @@ private void StartConnect()
...
+                    switch (_config.proxy.proxyType)
+                    {
+                        case ProxyConfig.PROXY_SOCKS5:
+                            remote = new Socks5Proxy();
+                            break;
+                        case ProxyConfig.PROXY_HTTP:
+                            remote = new HttpProxy();
+                            break;
+                        default:
+                            remote = new Socks5Proxy();
```
wongsyrone: How about throw an exception

PR #899, opened 2016-11-20
Assign SelectedIndex of ServerListBox only when it is range

hophacker:
Bug fix.
This could make shadowsocks crashed if _lastSelectedIndex is changed and not in range.


wongsyrone:
any procedure to reproduce the issue you intend to fix?

hophacker:
* Start shadowsocks
* Create a server config
* Leave shadowsocks
* empty "configs" in gui-config.json
* Start shadowsocks again.
* Crash

Noisyfox:
Fixed via commit 6df3250a9f766927e450df8769987cf92d572f6e

PR #917, opened 2016-11-30
Add optional tag support for ss url

wongsyrone:


wongsyrone:
ref: https://github.com/shadowsocks/shadowsocks-org/issues/25

PR #920, opened 2016-12-02
Fix privoxy listening port conflict

wongsyrone:
Let TCP stack find a free port for us


wongsyrone:
#910

PR #954, opened 2016-12-17
Fix #953

### LINKED ISSUE #953: Remove obsolete RefreshTrayArea() in PrivoxyRunner.cs
From: #950

Background: on Windows desktop, if a process has tray icon and get killed, the icon would be disappear only when user move the mouse pointer on it.

It was a hack to simulate the mouse move event (by User32.dll) on the tray area to clean up the killed Privoxy (was Polipo in previous versions) tray icon.

```
PrivoxyRunner.cs

        [DllImport("user32.dll")]
        public static extern IntPtr SendMessage(IntPtr hWnd, uint msg, int wParam, int lParam);

        private static void RefreshTrayArea(IntPtr windowHandle)
        {
            const uint wmMousemove = 0x0200;
            RECT rect;
            GetClientRect(windowHandle, out rect);
            for (var x = 0; x < rect.right; x += 5)
                for (var y = 0; y < rect.bottom; y += 5)
                    SendMessage(windowHandle, wmMousemove, 0, (y << 16) + x);
        }
```

However, it comes useless after @wongsyrone modified the Privoxy binary that Privoxy will never show an icon on the tray. Usage of Windows API` GetClientRect(windowHandle, out rect)` also get stuck when Shadowsocks-Windows is launched from `Sandboxie`.

Recommend to remove this function.

celeron533:
Remove obsolete RefreshTrayArea() in PrivoxyRunner.cs

### DIFF
--- a/shadowsocks-csharp/Controller/Service/PrivoxyRunner.cs
+++ b/shadowsocks-csharp/Controller/Service/PrivoxyRunner.cs
@@ -76,7 +76,6 @@ namespace Shadowsocks.Controller
                  */
                 PrivoxyJob.AddProcess(_process.Handle);
             }
-            RefreshTrayArea();
         }
 
         public void Stop()
@@ -87,7 +86,6 @@ namespace Shadowsocks.Controller
                 _process.Dispose();
                 _process = null;
             }
-            RefreshTrayArea();
         }
 
         private static void KillProcess(Process p)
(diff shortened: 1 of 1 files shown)


wongsyrone:
Done in https://github.com/shadowsocks/shadowsocks-windows/commit/1a283aa4ea46829def456bf28fc89cf34d275474

PR #1019, opened 2017-01-31
Chart Refactor

celeron533:
- Rename variable 'traffic' to 'trafficPerSecondQueue' and 'trafficInfoQueue'
- Remove class QueueLast<T> by using linq Queue.Last()
- Slight changes, such as rename, move several variables from function domain to class domain to improve performance. Logic remains the same.
- Update the logic of copying queue from Controller to View.


wongsyrone:
Looks good.

wongsyrone:
You should move `public class TrafficInfo` out of LogForm.cs. Otherwise, you cannot activate its Designer in VS.

PR #1023, opened 2017-02-03
Add an option to enable to using system temp folder.

hellofwy:
To solve issues like #1022.
Now users can choose to put temp files into system temp folder.

Only see the files changes in commit 4c49f1b73637d778dd76f76845f2f0c92dbbf6f8, which makes the  portable mode as default. 
Not read all codes, so may have bugs.
Build and run on my PC in x86 debug mode successfully.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,12 @@ if you want to deactivate all, please clear all textboxes.
 
 Please visit [Servers] for more information.
 
+#### Use System Temp Folder		
+		
+ If you want to put all temporary files into system temp folder instead of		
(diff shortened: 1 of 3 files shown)


celeron533:
So... revert back to its second last commit but keep the `use_system_temp_folder.txt` file?

crzidea:
I really need this feature!

wongsyrone:
Yes. You just revert back to the previous way.
Won't fix in the master branch. If you have such need, fork it yourself.

PR #1061, opened 2017-03-05
I18n Improvements

celeron533:
* Fix the unexpected empty lines of item [&View] in both Traditional Chinese and Japanese language definitions.
* Uniform the EOL (end of line) style. Convert CF (Unix) to CRLF (Windows) in both Traditional Chinese and Japanese definitions.
* Using a better method to initialize international languages.
* Rename language resources: cn -> CHS, zh-tw -> CHT, jp -> JPN


wongsyrone:
What's the benefit from this new method to initialize I18n?

celeron533:
@wongsyrone , I use MS standard 3 letters language code to keep the I18n logic simple. I also noticed that some of the items are defined as `ZZZ` which are not common used languages. So I think `ZZZ` is not a problem now.

wongsyrone:
IMHO the previous logic is simple enough.

hellofwy:
1. Making a standard language translation procedure is good.
2. I'm not family with c#'s language translation style, so I can't give any advice.
3. The [link][1] you provides don't give a full  list for all language.
4. Translation files can put into sub-folders if could.
5. I think there are too many changes in one commits.

[1]: https://msdn.microsoft.com/en-us/library/system.globalization.cultureinfo.threeletterwindowslanguagename(v=vs.110).aspx

wongsyrone:
To clarify my point, I18n initialization is done once at startup, The string value from EnglishName is short enough for string comparison, and the code readability is good enough even without comments.

There are no noticeable performance improvement with this new method.

celeron533:
You have convinced me.
> To clarify my point, I18n initialization is done once at startup, The string value from EnglishName is short enough for string comparison, and the code readability is good enough even without comments.
>There is no noticeable performance improvement with this new method.

PR #1083, opened 2017-03-22
Show save update file dialog instead of only opening temp folder

xJoeWoo:
Show a dialog to save downloaded update file. Freeing humans from copy-paste things.


wongsyrone:
I don't think this feature is really useful.

Issue #1063, opened 2017-03-06
退出Shadowsocks之后，IE的连接设置，依然留下127.0.0.1的PAC配置痕迹，建议自动备份恢复SS覆盖前PAC配置


Azlun:
Please read Wiki carefully, especially https://github.com/shadowsocks/shadowsocks-windows/wiki/Troubleshooting.
Please answer these questions before submitting your issue. Thanks!

### Version(release version or AppVeyor link)

3.4.3
### Environment(Operating system, .NET Framework, etc)
Win8.1, maybe .NET 4.2

### Steps you have tried
1.Start Shadowsocks.exe and using it very well in my office LAN;
2. Exit Shadowsocks.exe check IE Internet configuration about PAC configuration
3. I am expecting my old PAC from company is restored, 
     ticked:Auto detect PAC configuration;
     ticked:use configuration script: http://10.128.62.123:82/proxy.pac
    but the fact is I can see 
     ticked:Auto detect PAC configuration;
     unticked:use configuration script: http://127.0.0.1:1080/pac?t=20170306101753700&secret=+/Y3+2yXKzf5CfC5IerulJDklJ1MQinZ26BA8hqaZGA=

### What did you expect to see?
I am expecting my old PAC from company is restored, 
     ticked:Auto detect PAC configuration;
     ticked:use configuration script: http://10.128.62.123:82/proxy.pac
Internet should be OK for China sites.

### What did you see instead?
     ticked:Auto detect PAC configuration;
     unticked:use configuration script: http://127.0.0.1:1080/pac?t=20170306101753700&secret=+/Y3+2yXKzf5CfC5IerulJDklJ1MQinZ26BA8hqaZGA=
Internet are all fail"无法显示此页"
### Config and error log in detail (with all sensitive info masked)

wongsyrone:
If it doesn't affect anything, you can ignore it safely.

BTW, don't try to poke pac URL yourself, it is managed by SS itself.

Azlun:
I am blocked to use the network after exit ss,that is why I raise such a case. I guess most use with old PAC setting before as start will have such critical issue,since SS should manage to restored old setting but I my case not...

waitray:
i really have the same problem. if ss could restored my old pac setting,that is much better.

wongsyrone:
Done.

PR #1087, opened 2017-03-27
Restore user's proxy settings when ss proxy is disabled

xJoeWoo:
Saving user's proxy settings before ss service is enabled and restoring them on program exit.
Fix to #1063 #1078


wongsyrone:
NAK. You should submit patch to sysproxy project instead of poking registry directly.

wongsyrone:
https://github.com/shadowsocks/sysproxy/blob/master/sysproxy/main.c#L190

xJoeWoo:
Yea, that's the way far more elegant!

wongsyrone:
If you don't want to enable WPAD, just delete `PROXY_TYPE_AUTO_DETECT`

xJoeWoo:
Thanks! What about adding or editing a parameter for non-WPAD off? For example "off all" or something, it's compatible to old commands and offers new function.

wongsyrone:
I'm not sure whether ppl need WPAD or not. The users can disable it themselves via Internet Options in Internet Explorer.

xJoeWoo:
That's what my PR want to do, restoring user's settings :) I would submit to sysproxy project later, and a new commit in here about avoid writing to registry directly.

xJoeWoo:
I patched sysproxy project with two new usage to support this feature. (https://github.com/shadowsocks/sysproxy/pull/1](url)) Windows APIs are used in sysproxy.exe, they helps us to avoid poking registry in ss.
Also, I put record and restore operations into SystemProxy instead of ShadowsocksController. User's proxy settings will be restored as soon as user disable system proxy,

xJoeWoo:
I patched sysproxy project with two new usage to support this feature. https://github.com/shadowsocks/sysproxy/pull/1
Windows APIs are used in sysproxy.exe, they helps us to avoid poking registry in ss.
Also, I put record and restore operations into SystemProxy instead of ShadowsocksController. User's proxy settings will be restored as soon as user disable system proxy,

lixingcong:
Could you please tell me what is the meaning of NAK?

> NAK. You should submit patch to sysproxy project instead of poking registry directly.

wongsyrone:
Negative acknowledgment. In general, it is equivalent to reject.

wongsyrone:
Fixed via https://github.com/shadowsocks/shadowsocks-windows/commit/9ca61defe808efdf1e73966ac47009e54ca574f3

PR #1192, opened 2017-06-17
When system is shutting down, the proxy should be unset

huming2207:
Hello all,

I've found that when the system is shutting down, and shadowsocks app is killed but the proxy is still remain set improperly. When the system boot again later, the network is disconnected because Shadowsocks is not running. 

So the best way is to add a  [SystemEvents.SessionEnding](https://msdn.microsoft.com/en-us/library/microsoft.win32.systemevents.sessionending(v=vs.110).aspx) event and a event handler method to detect the user log out / shutdown behaviour and force unset the proxy when it happens.

Hope it helps anyway.

Regards,
Jackson

### DIFF
--- a/shadowsocks-csharp/Program.cs
+++ b/shadowsocks-csharp/Program.cs
@@ -51,6 +51,7 @@ namespace Shadowsocks
                 // handle non-UI exceptions
                 AppDomain.CurrentDomain.UnhandledException += CurrentDomain_UnhandledException;
                 Application.ApplicationExit += Application_ApplicationExit;
+                SystemEvents.SessionEnding += Application_ExitWhenSessionEnding;
                 SystemEvents.PowerModeChanged += SystemEvents_PowerModeChanged;
                 Application.EnableVisualStyles();
                 Application.SetCompatibleTextRenderingDefault(false);
@@ -182,6 +183,20 @@ namespace Shadowsocks
                 MainController.Stop();
                 MainController = null;
             }
+        }
+
+        private static void Application_ExitWhenSessionEnding(object sender, EventArgs e)
+        {
+            // Force disable proxy before exiting 
+            // We can't use Sysproxy here since it won't work when session is ending.
+            RegistryKey registryKey = Registry.CurrentUser.OpenSubKey("Software\\Microsoft\\Windows\\CurrentVersion\\Internet Settings", true);
+            registryKey.SetValue("ProxyEnable", 0);
+
+            // Remove other event in case something goes wrong in the next boot later on
+            Application.ApplicationExit -= Application_ApplicationExit;
+            SystemEvents.PowerModeChanged -= SystemEvents_PowerModeChanged;
+            Application.ThreadException -= Application_ThreadException;
+            HotKeys.Destroy();
         }
     }
 }
--- a/shadowsocks-csharp/Program.cs
+++ b/shadowsocks-csharp/Program.cs
@@ -189,8 +189,12 @@ namespace Shadowsocks
         {
             // Force disable proxy before exiting 
             // We can't use Sysproxy here since it won't work when session is ending.
-            RegistryKey registryKey = Registry.CurrentUser.OpenSubKey("Software\\Microsoft\\Windows\\CurrentVersion\\Internet Settings", true);
-            registryKey.SetValue("ProxyEnable", 0);
+            var proxyRegistryKey 
+                = Registry.CurrentUser.OpenSubKey("Software\\Microsoft\\Windows\\CurrentVersion\\Internet Settings", true);
+
+            // Disable proxy by setting ProxyEnable to 0 and deleting PAC.
+            proxyRegistryKey?.SetValue("ProxyEnable", 0);
+            proxyRegistryKey?.DeleteValue("AutoConfigURL");
 
             // Remove other event in case something goes wrong in the next boot later on
             Application.ApplicationExit -= Application_ApplicationExit;


chenshaoju:
The appveyor build is pass. (1.0.34)
But on Windows 10 1703, This function may not working properly.
After OS reboot,The proxy setting still there.

image

huming2207:
@chenshaoju  I should set both, including `ProxyEnable` and `AutoConfigURL`.

huming2207:
@chenshaoju Btw what do you think if we refactor all those proxy related classes (mainly for `Sysproxy` class) and use registry editing instead of calling `sysproxy.exe`,  just like I've done in this pull request? 

I think editing registry is more elegant, efficient and it will prevent some strange problems. For example, if we use `sysproxy.exe` to disable the proxy in session ending event handler, it will crash and keep the wrong proxy settings.

If you wish, I will help to rewrite and refactor the code.

wongsyrone:
> use registry editing

This is what we are trying to prevent.

Other ref: https://github.com/shadowsocks/shadowsocks-windows/pull/931

huming2207:
@wongsyrone Sorry I didn't quite understand. So basically...do you mean that if we just use WinINET or using "registry editing" will cause some other compatibility issues?

huming2207:
@wongsyrone But as I've said before, `Sysproxy` seems doesn't work when user is logging out, i.e.  when `SystemEvents.SessionEnding` event occurs. So probably we need to find some other ways to fix this bug.

wongsyrone:
Based on previous observation, wininet accepts native program only, e.g. 32-bit program on 32-bit OS, 64-bit program on 64-bit OS. This is why we add sysproxy.exe and sysproxy64.exe to the codebase since Shadowsocks.exe is compiled as 32-bit program.

SessionEnding is extremely unreliable, it doesn't get triggered as we expected. This situation applies to SystemEvents.PowerModeChanged as well.

To make life easier, I prefer to close SS manually before any system level event, e.g. reboot, shutdown, sleep and resume. I don't think it's a big issue for ordinary users. No matter what mode you are using, SS will restore user's wininet settings acquired by InternetQueryOption. I don't think there are many critical issues on proxy handling.

More info can be found in the issue tracker and PRs.

Some ref:
https://stackoverflow.com/questions/19476646/detect-restart-in-windows-8-sessionending-not-called
https://stackoverflow.com/questions/7678832/systemevents-sessionending-not-firing

huming2207:
@wongsyrone Ah okay, now I've got it, thanks. So I'd better figure out why `sysproxy` doesn't work when the event triggered and I'll try if I can solve it.

Meanwhile for the event itself, as you said, what if we use `WM_EndSession`? Will it be better instead of using `SessionEnding` event?

wongsyrone:
I didn't try WM_EndSession. At that time, I guess system will prevent any further call and response to user's request ASAP. It might be the root cause of malfunction of SessionEnding and Sysproxy.exe.

huming2207:
Hmm yea, maybe the system just prevents those process calls. That would be very frustrating. 

But...I'm sorry that I'm still wondering why we can't edit registry.  It doesn't call any WinINET API directly, so I think it should works both in 32 bit and 64 bit system.  Or maybe I should test again on some 32 bit systems later on in my virtual machine lol.

wongsyrone:
> so I think it should works both in 32 bit and 64 bit system.

Don't forget to try it on Win 10 Insider Preview, where things getting messy

huming2207:
@wongsyrone Haha yea I'll try it later next week when I finish my final exam.

huming2207:
@wongsyrone @chenshaoju Hi guys,

Sorry about that I've totally forgot this issue after finish the exam lol...

Just proved that my patch does work in Windows 10 16193, no issues at all. 

Firstly, I simply run my patched shadowsocks client on my virtual machine, worked fine...
test1

**Then later I rebooted the VM directly, without correctly exiting the shadowsocks client, and global proxy has switched off as it should have been.**
test2

So in my opinion, we should remove Sysproxy and use registry editing. It is proved to be safe and meanwhile it can reduce the executable file size.

Regards,
J

huming2207:
@chenshaoju @wongsyrone Btw editing these values also documented in Microsoft's official document and blog. So I think it should be safe enough and it is better than Sysproxy anyway.

See here: https://support.microsoft.com/en-us/help/819961/how-to-configure-client-proxy-server-settings-by-using-a-registry-file

https://blogs.msdn.microsoft.com/askie/2015/10/12/how-to-configure-proxy-settings-for-ie10-and-ie11-as-iem-is-not-available/

huming2207:
Ah sorry I've forgot that editing can be done but it can't be refreshed. 

See here: https://stackoverflow.com/questions/2020363/how-to-change-global-windows-proxy-using-c-sharp-net-with-immediate-effect

As a result, Sysproxy method should be reserved for setting the proxy, while setting registry by my `Application_ExitWhenSessionEnding` event is still usable since it just remove the wrong setting before system session ends (which will be refreshed later when it boots again).

huming2207:
Now the PAC issue mentioned by @chenshaoju was fixed, tested on Windows 10 15063.483 and Insider 16193.1001.  Please have a try and see if it works in your environment.

huming2207:
@chenshaoju @wongsyrone Any updates?

LiteCoder:
agree.i like this feature.

chenshaoju:
Sorry for long time,Busy like hell. The test has passed.

image

huming2207:
@chenshaoju Thanks for your reply. I was thinking you've got write permission to this repo so I @ you to confirm and merge lol...Anyway let's wait someone else to test my code and merge it later if it really works.

huming2207:
@celeron533 大神你好，方便的话能看下这个pull request吗？如果可以的话就merge吧，不行就close得了，拖太久了……

chenshaoju:
@wongsyrone

wongsyrone:
现在这个项目由 celeron533 维护

huming2207:
好的明白，谢谢，辛苦了。
@celeron533 大佬求关照……

celeron533:
@huming2207 , thank you for your work.
The code implemented the Proxy Server reset function and is able to solve some issue tickets.
Unfortunately, I could not merge it now for the following reasons:
1. The implementation just disable the proxy and wipe up the AutoConfigURL.
If the user already have some specific Proxy settings before launching ss-win, commit 816651e (Save user wininet settings as another config file) do the backup & restore. However, this pull request's design remove everything. Yes, you could read the proxy backup file and write it back to registry, but it increases the code and logic complexity.
```
proxyRegistryKey?.SetValue("ProxyEnable", 0);
proxyRegistryKey?.DeleteValue("AutoConfigURL");
```
image

2. I knew there are some limitations about the wininet library and you would like to use registry editing.
As @wongsyrone mentioned
>This is what we are trying to prevent.
Other ref: #931

Considering `sysproxy` is inspired from https://github.com/getlantern/lantern/issues/5346 , they may have solved this issue when shutting down the Windows.

huming2207:
@celeron533 Thanks for your reply.

For the second part the limitation doesn't actually exist if we modify registry, at least I didn't find any further issues in Windows 10 Insider as I've mentioned in this pull request some days before. I'll do it  again with some newer Windows 10 insider versions if necessary.

For the first part as you said, I'll make some changes to improve those backup/restore logics later if possible.

Quite a lot of users were annoyed by this issue. So I think it is reasonable to implement something to fix it.

celeron533:
There is an interesting phenomenon on my working PC.
The working PC installed ForcePoint Internet access management system. It locks the Internet settings. But the Sysprocy overcomes it...

celeron533:
`sysproxy` implementation from https://github.com/getlantern/sysproxy-cmd

`sysproxy wait-and-cleanup` differs from `sysproxy off` in that it waits for input from stdin (or close) before turning off proxy setting. Any signal or Windows system shutdown message triggers the cleanup too.

https://github.com/getlantern/sysproxy-cmd/blob/master/main.c#L21
https://github.com/getlantern/sysproxy-cmd/blob/master/main.c#L57

huming2207:
@celeron533 Sorry mate got too busy these days and I forgot to reply you.

Do you mean that if we add a ```sysproxy wait-and-cleanup``` instead of ```sysproxy off``` and it should works as we expected (i.e. triggered by the signals)?

celeron533:
@huming2207 Perhaps need to refactor the embedded `sysproxy.exe` to `sysproxy.dll` bcz Windows may not allow to start new program during logging off progress.

wongsyrone:
lantern对sysproxy的处理可以达到要求，但是很容易被安全软件报告为病毒。

huming2207:
@wongsyrone 抱歉这个年代有点久远……想问下你的意思是说sysproxy自己负责重置这个代理设置还是lantern对它做了啥手脚可以实现这种功能？

celeron533:
Currently sysproxy.exe is only executed when there is a command need to be executed such as `query`, `set`, etc. When Windows is shutting down, it has difficulty to call sysproxy again in the shutdown process.
A possible solution is to let sysproxy run as an individual background daemon and using named pipe to send commands on the fly. Then do same thing as `lantern` did in setupSystemShutdownHandler() to watch  system shutdown event.

studentmain:
any update?

lhy0403:
I personally don’t like embedding binary executables in the codebase. I’m also curious on why direct registry access in not a proper choice.
It’s historical thread please merge or close it ASAP.

studentmain:
As a part of x64/Any CPU support project, I bring back WinINet to v5(net-core) branch to drop sysproxy, that would resolve this problem.

I think we needn't support windows insider preview version, they're not stable at all. Some test shows the original problem which make us replace WinINet with Sysproxy not exist in windows stable version.

studentmain:
#2874 Should resolved this problem.

PR #1232, opened 2017-07-12
Fixing issues with "4.0.4 null exceptions" at launch for certain users

Chopinsky:
Adding rail guard and fallback logic in the initializing logic, where if a user setting is previously null, we could end up with null reference exceptions being thrown.

### DIFF
--- a/shadowsocks-csharp/Util/SystemProxy/Sysproxy.cs
+++ b/shadowsocks-csharp/Util/SystemProxy/Sysproxy.cs
@@ -48,7 +48,12 @@ namespace Shadowsocks.Util.SystemProxy
         public static void SetIEProxy(bool enable, bool global, string proxyServer, string pacURL)
         {
             Read();
-            if (_userSettings == null || !_userSettings.UserSettingsRecorded)
+            if (_userSettings == null)
+            {
+              _userSettings = new SysproxyConfig();
+            }
+
+            if (!_userSettings.UserSettingsRecorded)
             {
                 // record user settings
                 ExecSysproxy("query");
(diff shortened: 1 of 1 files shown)


wongsyrone:
不用这么麻烦，把FileNotFoundException改成Exception就行了，问题出在反序列化上

wongsyrone:
之前的策略是只有当找不到文件时才丢弃之前的设置，为了避免麻烦可以改成不管是找不到文件，还是反序列化失败，都丢弃之前的设置。总之有得必有失

Chopinsky:
Still need to take care of the cases where no exception is thrown but deserializer returns null -- the caller has assumed that after calling Read(), `_userSettings` should have be initialized, which is not always true if we don't implement the `finally` code block.

wongsyrone:
改成这样就行了

```
        private static void Read()
        {
            try {
                string configContent = File.ReadAllText(_userWininetConfigFile);
                _userSettings = JsonConvert.DeserializeObject<SysproxyConfig>(configContent);
            } catch (Exception) {
                // suppress all exceptions
            } finally {
                if (_userSettings == null) _userSettings = new SysproxyConfig();
            }
        }
```

celeron533:
这样的话，如果ss-win的持久化文件有重大升级变更可能会导致老版本数据丢失，好在`user-wininet.json`也不是很重要的东西
> 之前的策略是只有当找不到文件时才丢弃之前的设置，为了避免麻烦可以改成不管是找不到文件，还是反序列化失败，都丢弃之前的设置。总之有得必有失

wongsyrone:
> 如果ss-win的持久化文件有重大升级变更

其实不出意外的话是不需要其他信息的，如果M$不改API的话

celeron533:
Good idea
>改成这样就行了
```
        private static void Read()
        {
            try {
                string configContent = File.ReadAllText(_userWininetConfigFile);
                _userSettings = JsonConvert.DeserializeObject<SysproxyConfig>(configContent);
            } catch (Exception) {
                // suppress all exceptions
            } finally {
                if (_userSettings == null) _userSettings = new SysproxyConfig();
            }
        }
```

Chopinsky:
Updated

celeron533:
Thanks @Chopinsky 
Squashed and merged this commit in f0b196c8c305f2e5ae1be551a6406a6af15c1473

PR #1285, opened 2017-08-01
Some minor refector work

patrioticwinnie:
Just warm up in studying the code, want to contribute some help to break the wall. I can start with doing some refector work.

### DIFF
--- a/shadowsocks-csharp/Proxy/Socks5Proxy.cs
+++ b/shadowsocks-csharp/Proxy/Socks5Proxy.cs
@@ -43,6 +43,7 @@ namespace Shadowsocks.Proxy
 
         private const int Socks5PktMaxSize = 4 + 16 + 2;
         private readonly byte[] _receiveBuffer = new byte[Socks5PktMaxSize];
+        private static readonly byte[] Handshake = {5, 1, 0};
 
         public EndPoint LocalEndPoint => _remote.LocalEndPoint;
         public EndPoint ProxyEndPoint { get; private set; }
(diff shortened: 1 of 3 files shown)


wongsyrone:
1. StringEx comes from NuGet, send patch to that project
2. MemoryStream should be disposed, read MSDN carefully: https://msdn.microsoft.com/en-us/library/system.io.memorystream(v=vs.110).aspx

> This type implements the IDisposable interface. When you have finished using the type, you should dispose of it either directly or indirectly. To dispose of the type directly, call its Dispose method in a try/catch block. To dispose of it indirectly, use a language construct such as using (in C#) or Using (in Visual Basic). For more information, see the “Using an Object that Implements IDisposable” section in the IDisposable interface topic.

patrioticwinnie:
Aye aye captain, modified as your advice.

wongsyrone:
BTW, `new MemoryStream(buf)` in `GZipStream input = new GZipStream(new MemoryStream(buf)` might cause memleak. The ref got lost without disposing. Please correct me if I'm wrong.

I'm not the captain, @celeron533 is.

patrioticwinnie:
Actually, missing dispose a MemoryStream, especially a local MemoryStream won't mess up anything. If you reflector the source code, you'll find MemoryStream's Dispose method only set some flag to false state. According to the discussion (https://stackoverflow.com/questions/4274590/memorystream-close-or-memorystream-dispose?answertab=votes )
The Dispose method is designed to release unmanaged resources, missing calling Dispose which cause either resource leak or high pressure on Finalizer thread. In this case, you can ignore this local MemoryStream. Relax.

PR #1299, opened 2017-08-06
fixed LoadConfiguration parameter not use bug

0x9c:


celeron533:
It is caused by #1301

celeron533:
Fixed in 081bbaf011374181058a17080c84fd75a9eb67c6

PR #1298, opened 2017-08-06
SIP002 and SIP003 support.

rwasef1830:
Hello,
This is a first draft WIP to get feedback on my approach. This pull request implements SIP002 url (parsing and generation for QR code with backward-compatibility of accepting old-style urls and generating old-style urls when there is no plugin for the config), and SIP003 plugin interface for shadowsocks-windows.

- SIP002 has unit tests and proven to work.
- SIP003 needs a little bit of testing, would appreciate feedback on the approach and fresh eyes to see if I made any mistakes.

Hope to eventually get this merged soon :-)

Issue #1045, opened 2017-02-22
Enable TCP Fast Open for Windows


icylogic:
>But even without TLS 1.3, we can combine TCP Fast Open and the TLS False Start option, and reduce the delay from 3-RTT to 1-RTT.
![](https://winblogs.azureedge.net/win/2016/06/TFS-TFO-Diagram-1024x562.png)
https://blogs.windows.com/msedgedev/2016/06/15/building-a-faster-and-more-secure-web-with-tcp-fast-open-tls-false-start-and-tls-1-3/#ebufEVtM1OLzRVuc.97

TCP Fast Open on Windows (Anniversary Update):

```C#
var s = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
var e = new SocketAsyncEventArgs();
s.SetSocketOption(SocketOptionLevel.Tcp, (SocketOptionName)15, true);
e.SetBuffer(buffer, offset, length);
s.ConnectAsync(e);
```

Values for socket options are defined in `10.0.14393.0\shared\ws2ipdef.h`

https://msdn.microsoft.com/en-us/library/windows/desktop/ms738596(v=vs.85).aspx

After receiving the TFO Cookie, we could send SYN+Data with Cookie:

![2017-02-22_09-29-57](https://cloud.githubusercontent.com/assets/2814357/23215799/a0c56ddc-f8e1-11e6-9a2b-05eea0b23e83.png)

TFO has been available for shadowsocks servers on linux for years. I tried to introduce this feature into shadowsocks windows, but it seems to be a big change for the most critical parts (TCPRelay, Socks5Proxy, WrappedSocket, etc.) So I put information here and leave it for you to decide.

pexcn:
win7也支持吗？

icylogic:
@pexcn 
![2017-02-22_10-04-01](https://cloud.githubusercontent.com/assets/2814357/23217148/4a4691fc-f8e6-11e6-889a-16e11c093c34.png)

pexcn:
哦哦，原来如此…

wongsyrone:
The best way to do so is to create another x86 build profile and use the preprocessor to distinguish different profiles. For TFO variant, one should check OS when the program startup.

icylogic:
@wongsyrone That would definitely work. 
[...]

PR #1328, opened 2017-08-19
Add TCP Fast Open support

wongsyrone:
Forward proxy has been removed since we cannot control
its socket options.


wongsyrone:
It hasn't been tested on Windows 7; I don't have such devices.

chenshaoju:
The Appveyor url in readme.md is broken ( https://ci.appveyor.com/project/wongsyrone/shadowsocks-windows-yqdou ) Unable test.

(Too lazy for compile 😅 )

wongsyrone:
https://ci.appveyor.com/project/celeron533/shadowsocks-windows/build/4.0.5.24

chenshaoju:
On Windows 7,Log print:
```
[2017-08-19 17:08:48] TFO: False
```

wongsyrone:
It's expected. https://github.com/shadowsocks/shadowsocks-windows/issues/1045

celeron533:
The requirement is conflick with pull request #1214 (Forward http proxy with Auth) :worried:
Also may have impact with #1298 (SIP002, SIP003)

wongsyrone:
I talked to @madeye before, he agreed to remove Forward proxy.

madeye:
Is it possible to implement forward proxy as a SIP003 plugin instead? If so, we can keep our core code base as clean as possible.

madeye:
IMO, the priority list of pull requests is SIP003 > TFO > Forward proxy.

wongsyrone:
Okay. This will remains a reference implementation for people who interested on both of them. I choose not to modify my PR.

rwasef1830:
@madeye @wongsyrone  Why is this in conflict with my pull request ? I don't touch any of the core tcp / udp relay code (I implemented it in a far simpler way from a higher level).

I see no reason why we can't have tcp fast open at least for the supported cases (no forward proxy, no sip003 plugin).

madeye:
Yes, if there is no conflict, I think we can try to merge this PR next.

@rwasef1830 Any interest to help merge this one?

rwasef1830:
Yes, should I open a new pull request and cherry pick the commit to my fork ? I want to maintain credits to @wongsyrone

madeye:
I think we'd better open a new pull request and rebase the latest changes then.

@wongsyrone What do you think?

wongsyrone:
There are two must-have for TFO.

1. be able to control remote socket option
2. be able to receive handshaking from client thus we have data to send when connecting

Under forward proxy, I have no idea on fulfilling these two requirements.

Currently, I'm investigating c-ares, cannot find much time to rebase. If someone has interest, he can mention my ID in the commit message. As I said before, it serves as a reference implementation, everyone can use without notice.

rwasef1830:
@wongsyrone ok, thanks. I will cherry pick and open a new pull request.

celeron533:
Sorry I have not get sufficient time to review all the code of this pull request and do a full assessment of code impacts.

wongsyrone:
Forgot to mention, the two must-have above is specific to the Socket.ConnectAsync() way. If you want to pass parameters to the plugin like: "param=tcp-fast-open;blah-blah" to notify other layers to enable TFO. That layer should use ConnectEx on Windows, otherwise, it won't work.

rwasef1830:
@wongsyrone Yeah I noticed this. Currently, I am trying to make the changes as small as possible to implement this, but I really have the urge of rewriting that layer.. The ancient BeginXXX EndXXX style code is very messy and hard to follow.

wongsyrone:
Do you want to use TAP? I already implemented it with SAEA for the whole networking layer.

rwasef1830:
I want to change IProxy interface to expose TAP and wrap SAEA into TAP interface, it is much cleaner. Even better if I can make it expose Stream interface, so unit testing becomes possible.

wongsyrone:
In fact, there is no need to convert all code into TAP. If you use SAEA, there is no need to keep connect timer, just let the socket layer handle it.

PR #1336, opened 2017-08-23
Use job objects with SIP003 plugin to avoid lingering processes if app is killed

rwasef1830:
This is a follow up to #1298. Uses windows job objects with SIP003 processes so that if shadowsocks is killed SIP003 plugin is killed by OS also (instead of lingering processes).

Hoping for a swift review and merge :-)

### DIFF
--- a/shadowsocks-csharp/Controller/Service/Sip003Plugin.cs
+++ b/shadowsocks-csharp/Controller/Service/Sip003Plugin.cs
@@ -5,6 +5,7 @@ using System.Net;
 using System.Net.Sockets;
 using System.Reflection;
 using Shadowsocks.Model;
+using Shadowsocks.Util.ProcessManagement;
 
 namespace Shadowsocks.Controller.Service
 {
@@ -15,6 +16,7 @@ namespace Shadowsocks.Controller.Service
         public int ProcessId => _started ? _pluginProcess.Id : 0;
 
         private readonly object _startProcessLock = new object();
+        private readonly Job _pluginJob;
         private readonly Process _pluginProcess;
         private bool _started;
         private bool _disposed;
@@ -66,6 +68,8 @@ namespace Shadowsocks.Controller.Service
                     }
                 }
             };
+
+            _pluginJob = new Job();
         }
 
         public bool StartIfNeeded()
@@ -88,6 +92,7 @@ namespace Shadowsocks.Controller.Service
                 _pluginProcess.StartInfo.Environment["SS_LOCAL_HOST"] = LocalEndPoint.Address.ToString();
                 _pluginProcess.StartInfo.Environment["SS_LOCAL_PORT"] = LocalEndPoint.Port.ToString();
                 _pluginProcess.Start();
+                _pluginJob.AddProcess(_pluginProcess.Handle);
                 _started = true;
             }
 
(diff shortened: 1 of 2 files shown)


celeron533:
Appreciate for the pull request @rwasef1830 

Tested with taskkill via taskmanager. Looks good.

------

Found a defect from previous Pull Request:
Plugin does not restart properly when user change the server settings.

In file `ShadowsocksController.cs`
In `Reload()` method, there is a `StopPlugins()`, but there is no `StartPlugins()`.
Scenario:
When user change the plugin or the plugin options on the Config Form then click OK, the plugin process would be terminated but never be restarted (meanwhile the Server's IP, port, password, encryption are applied with the latest configuration).

-----

Another suggestion for usability:
When loading a plugin, record the load event to log file.
Example:
When everything is okay:
```
[timestamp] Shadowsocks started
[timestamp] Loaded Plugin: foo.exe. Option: bar
```
When the plugin is not found:
```
[timestamp] Shadowsocks started
[timestamp] Failed to load Plugin: foo.exe. Option: bar
```

rwasef1830:
@celeron533 
> Found a defect from previous Pull Request:
Plugin does not restart properly when user change the server settings.
> In file ShadowsocksController.cs
In Reload() method, there is a StopPlugins(), but there is no StartPlugins().
> Scenario:
> When user change the plugin or the plugin options on the Config Form then click OK, the plugin process would be terminated but never be restarted (meanwhile the Server's IP, port, password, encryption are applied with the latest configuration).

This is intentional plugin is started on first connection (see usage of ```Sip003Plugin.StartIfNeeded()````), each tcp connection, sip plugin is checked if running or not and process is started (if killed).

I tested it and it seems to work on my machine fine. If it is failing with you, I would appreciate a reliable case to reproduce the defect so I can fix it.

> Another suggestion for usability ...

Great idea! I will update this pull request with the changes.

celeron533:
@rwasef1830 
Just for discussion.
The design is 'start plugin on demand'. If it cost some time to initialize the plugin. When the client needs to send the 1st package, it would be delayed.

celeron533:
If the plugin is enabled, should it disable Forward Proxy function?
image

rwasef1830:
@celeron533 You make a fair point. I just pushed a commit to early start SIP003 plugins to avoid delay on first packet.

Already when SIP003 is enabled, forward proxy is ignored (but I don't disable it globally, if another server doesn't have SIP003 plugin, it will be used normally).

I considered updating the forward proxy UI checkbox to say ```(doesn't apply to servers with plugin)``` but I can't update the translations so I didn't do it.

rwasef1830:
@celeron533 Anything blocking this from merge ?

celeron533:
Hi @rwasef1830 , I asked some users and few of them have more than 20 servers. Perhaps the original `plugin.StartIfNeeded()` (late load) is more suitable for these scenarios. Sorry for the confusion :(

In the coming 3 weeks, I would be busy and may not reply or check the pull request on time.

rwasef1830:
@celeron533 Even if 20 servers, early start would not be too bad, it is not really connecting to server, just launching the plugin process. This shouldn't take too long on any recent PC. I think it is fine with early start.

PR #1540, opened 2017-11-29
make ss support kuntun

DaPangR:
kcptun can import ss network

 speed, but ss plugins function cannot use kcp now. So i modify a few code, make sure ss plugin can use kcptun.


pexcn:
https://github.com/shadowsocks/kcptun/releases/tag/v20170718

celeron533:
You destroyed the standard SIP003 implementation.
You also remove the description in readme.md

PR #1560, opened 2017-12-06
Simplification of Util.CS

Lightveil:
Simplified the code of FormatBytes and GetBandwidthScale in Util.CS.


celeron533:
Thanks for sending the pull request.
IMO, the updated logic becomes hard to understand and there is no performance benefits from lambda expressions.

Issue #1672, opened 2018-02-10
Add an option to let user wrap SIP003 environment variables as normal in-line arguments


celeron533:
### Just a brainstorm

According to https://github.com/shadowsocks/shadowsocks-org/issues/28 , the SIP003 arguments are delivered by environment variables:
Mandatory:
`SS_REMOTE_HOST`, `SS_REMOTE_PORT`, `SS_LOCAL_HOST` and `SS_LOCAL_PORT`
Optional:
`SS_PLUGIN_OPTIONS`

--------
Currently, a SIP003 "compatible" plugin program must able to read and grab its Process environment variables when being launched. Equivalent to the following console commands:

1. ss-win set the environment variables
```
set SS_REMOTE_HOST=1.2.3.4
set SS_REMOTE_PORT=9999
set SS_LOCAL_HOST=127.0.0.2
set SS_LOCAL_PORT=33333
set SS_PLUGIN_OPTIONS=foobar
```
2. Then ss-win launch the plugin
```
start someplugin.exe
```

--------
If user has plugins whose implementation does not following SIP003 standard, user should have a work-around to give the in-line arguments when starting the plugin program. For instance:

1. ss-win set the environment variables
```
set SS_REMOTE_HOST=1.2.3.4
set SS_REMOTE_PORT=9999
set SS_LOCAL_HOST=127.0.0.2
set SS_LOCAL_PORT=33333
set SS_PLUGIN_OPTIONS=foobar
```
2. User typed the format and pattern for the in-line arguments wrapper
```
-remote %SS_REMOTE_HOST%:%SS_REMOTE_PORT% -local %SS_LOCAL_HOST%:%SS_LOCAL_PORT% %SS_PLUGIN_OPTIONS%
```
3. Then ss-win launch the plugin by using the wrapper
```
start someplugin.exe -remote %SS_REMOTE_HOST%:%SS_REMOTE_PORT% -local %SS_LOCAL_HOST%:%SS_LOCAL_PORT% %SS_PLUGIN_OPTIONS%
```
which will be parsed into
```
start someplugin.exe -remote 1.2.3.4:9999 -local 127.0.0.2:33333 foobar
```

It should expand the selection of third-party plug-in programs.

madeye:
The reason we use environment variable is to avoid special character issue in the command line, e.g key s like `^%$&^\/`.

However, it's still worth to support CLI args in windows as some users would try to use non-SIP003 plugins.

PR #1684, opened 2018-02-17
1. Pass options to SIP003 plug-in through the Agruments property; 2. …

ylhyh:
…The Port of SIP003's "LocalEndPoint" should be Remote Server's port when a plug-in enabled.

### DIFF
--- a/shadowsocks-csharp/Controller/Service/Sip003Plugin.cs
+++ b/shadowsocks-csharp/Controller/Service/Sip003Plugin.cs
@@ -55,6 +55,7 @@ namespace Shadowsocks.Controller.Service
                 StartInfo = new ProcessStartInfo
                 {
                     FileName = plugin,
+                    Arguments = pluginOpts,
                     UseShellExecute = false,
                     CreateNoWindow = true,
                     ErrorDialog = false,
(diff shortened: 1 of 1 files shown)


shadowsocks-csharp/Controller/Service/Sip003Plugin.cs
```diff
@@ -86,7 +87,7 @@ public bool StartIfNeeded()
                     return false;
                 }
 
-                var localPort = GetNextFreeTcpPort();
+                var localPort = int.Parse(_pluginProcess.StartInfo.Environment["SS_REMOTE_PORT"]);
```
celeron533: `SS_LOCAL_HOST` `SS_LOCAL_PORT`: the plugin listens SS native connection.
`SS_REMOTE_HOST` `SS_REMOTE_PORT`: the plugin then connect to the remote server.

SS ===> to plugin's LOCAL HOST/PORT ===> plug-in data processing ===> send to REMOTE HOST/PORT

ylhyh:
@celeron533 , I agree with you that Plugin should following the standard,

celeron533:
@ylhyh ,
 Both designs has their pros and cons.
`GetNextFreeTcpPort()` makes sure that the port has no conflict and free to use.
`SS_REMOTE_PORT` (or `"SS_LOCAL_PORT"`) lock the port as pre-defined.

BTW, in your code change it is `SS_REMOTE_PORT` but in your comment it is `variable %SS_LOCAL_PORT%`. Which one do you want?

SIP003 is a standard. You 'should' always uses the plugin programs which follow the standard.
#1672 is talking about mapping the environment variables to parameter to give maximum compatibility to normal plugin programs.

ylhyh:
1. It seems %SS_REMOTE_PORT% is useless when a plugin being launched, so I use it to specify the plugin's listening port.
2. Yes, I did same thing as #1672 mentioned.

ylhyh:
anyway, cancel the request now.

PR #1732, opened 2018-03-11
add official way to dump network tracing info

wongsyrone:
just uncomment the specific node in app.config, no need to recompile
to take effect. It will be shipped as Shadowsocks.exe.config after built.

This is for debugging only.


celeron533:
Thanks @wongsyrone, it would be very useful for diagnosis

PR #1753, opened 2018-03-16
"Allow Clients from LAN" Description & Translation optimization

chenshaoju:
In #1740, I think this feature description may not clear.

This function actually binding the local socks5 proxy port on 0.0.0.0, In some public IP case, May cause undesired result.


shadowsocks-csharp/Data/zh_CN.txt
```diff
@@ -13,7 +13,7 @@ Edit Servers...=编辑服务器...
 Statistics Config...=统计配置...
 Start on Boot=开机启动
 Forward Proxy...=正向代理设置...
-Allow Clients from LAN=允许来自局域网的连接
+Allow other Devices to connect=允許其他裝置連入
```
celeron533: zh_TW

chenshaoju: Oops, My fault, Fixed.

PR #1818, opened 2018-04-17
Add plugin CLI arguments support

celeron533:
per #1672 

image


celeron533:
A simple way to test whether it work or not:
1. Type `cmd` as the plugin program
2. Type anything as the plugin options
3. Type `/k echo %SS_LOCAL_HOST% %SS_LOCSL_PORT% %SS_REMOTE_HOST% %SS_REMOTE_PORT% %SS_PLUGIN_OPTIONS%> output.txt` as the plugin arguments
4. Save changes. It will call `cmd.exe` to writ the environment variables to file `output.txt`, which may looks like:
```
127.0.0.1 5122 11.22.33.44 1234 pluginoptions
```

chenshaoju:
I'm not sure,  but this looks may have a security issue.
In extreme case, may cause the dangerous result like delete all files from the root or running script to download malware etc...

celeron533:
**Advantages**

User could use any plugin which does not follow the SIP003 standard. Currently only a limited number of SIP003 plugins available and some of them are out-of-date from their upstream release.


**Disadvantages**

As @chenshaoju mentioned, 
>this looks may have a security issue.

IMO, execute the plugin program already has a kind of risk.

chenshaoju:
These environment variables only pass by cmd.exe (or shell?)

If use some 3rd non-SIP003 plugin, may not work.

Directly give these environment variables name to the plugin, Most is not working.
image

But if pass by cmd.exe, The cmd.exe will give environment variables value to the plugin, It's will work.
image

It's possible to start cmd.exe first, and give these environment variables to cmd.exe, Let cmd.ext give these value to the plugin?

Or just direct give these value to the plugin in the command line.

https://github.com/shadowsocks/shadowsocks-windows/issues/1971

PR #1973, opened 2018-08-18
简化CompareVersion

TheCGDF:
简化CompareVersion


SoDa-GitHub:
https://docs.microsoft.com/en-us/dotnet/api/system.version.compareto?redirectedfrom=MSDN&view=netframework-4.7.2#System_Version_CompareTo_System_Version_

in c#, Version 1.1 is older than version 1.1.0.

so, it cannot pass the unitTest.

PR #2123, opened 2018-12-17
support serverConfigFile in gui-config.json

Streamlet:
Hi, I separated server config file from gui-config.json, by a hidden config item named "serverConfigFile".

I wrote so ugly code that you may re-write the feature. I hope you will consider this feature seriously.
^_^


celeron533:
But why you need to separate the config?

PR #2126, opened 2018-12-18
支持配置文件存放在单独的目录

Streamlet:
可在程序目录创建一个文件 path，内容为一个路径，这样就会从这个目录读取配置。
方便使用各种同步盘同步配置

### DIFF
--- a/shadowsocks-csharp/Model/ConfigPath.cs
+++ b/shadowsocks-csharp/Model/ConfigPath.cs
@@ -1,0 +1,33 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+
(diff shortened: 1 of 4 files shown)


Streamlet:
> Personally I prefer to add a new program startup argument as same as `libev` and `python` version, such as `-c <config_file>`

if current process of client was launched with `-c <config_file>`, and than set autorun in the UI, I wonder whether the next launching will be along with `-c <config_file>` or not?

studentmain:
> I wonder whether the next launching will be along with -c <config_file> or not?

It won't be problem:

图片

database64128:
This is overengineering for a desktop GUI client.

PR #2155, opened 2019-01-12
bug fix for missing data

ihciah:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

---

It may lost data when receiving the first packet from remote. For example, using stream cipher, if the length of the first packet received is shorter than IV, it will be discarded.

### DIFF
--- a/shadowsocks-csharp/Controller/Service/TCPRelay.cs
+++ b/shadowsocks-csharp/Controller/Service/TCPRelay.cs
@@ -831,7 +831,7 @@ namespace Shadowsocks.Controller
                     {
                         // need more to decrypt
                         Logging.Debug("Need more to decrypt");
-                        session.Remote.BeginReceive(_remoteRecvBuffer, 0, RecvSize, SocketFlags.None,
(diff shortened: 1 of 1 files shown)


ihciah:
I found the fix cannot fix the problem...
Will open another pr soon.

wongsyrone:
> if the length of the first packet received is shorter than IV, it will be discarded.

They are stored in the internal circular buffer, no matter what kind of method you are using, stream or AEAD. We apply greedy strategy to handle AEAD decryption.

ihciah:
> > if the length of the first packet received is shorter than IV, it will be discarded.
> 
> They are stored in the internal circular buffer, no matter what kind of method you are using, stream or AEAD. We apply greedy strategy to handle AEAD decryption.

I don't understand why not saving the received data if it's too short.
For example, if the data arrived byte-by-byte(one cannot promise the tcp data recv as send), they will all be discarded.
I think the data should be saved and appended, until the length is enough for decrypting.

Thanks for your reply.

ihciah:
> > if the length of the first packet received is shorter than IV, it will be discarded.
> 
> They are stored in the internal circular buffer, no matter what kind of method you are using, stream or AEAD. We apply greedy strategy to handle AEAD decryption.

I got it.
Thanks for your reply.

PR #2511, opened 2019-08-17
Added capability of specifying nicknames of the servers

lyr316:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature

---

Added capability of specifying nicknames of the servers
Explanation of your *pull request* in arbitrary form goes here. Please make sure the description explains the purpose and effect of your *pull request* and is worded well enough to be understood. Provide as much context and examples as possible.
Currently after a server is added, the server list would show the server address as an indicator of the server entry. However, an IP address is not a very user friendly indicator for human beings, especially when you have added multiple server entries.
This usability improvement would allow you to specify an optional nickname of the server entry. For example, you could give nickname "Server 1" to server "xxx.xxx.xxx.xxx". If the nickname is specified, it will be displayed in the server list and system tray menu, instead of the actual server address.

### DIFF
--- a/shadowsocks-csharp/Model/Server.cs
+++ b/shadowsocks-csharp/Model/Server.cs
@@ -29,6 +29,7 @@ namespace Shadowsocks.Model
         public string plugin_args;
         public string remarks;
         public int timeout;
+        public string nick_name;
 
         public override int GetHashCode()
         {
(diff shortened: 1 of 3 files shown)


celeron533:
We already has similar field called 'remarks'

PR #2695, opened 2019-11-24
support scan ssr QrCode;

pedoc:

- [ ] Bug fix
- [x] Improvement
- [x] New feature

---

support scan ssr QrCode;
copy menu item 'scan QrCode from screen' the top;


celeron533:
1. `Fody is only supported on MSBuild 16 and above. Current version: 15.`
2. The "additional" SSR features are not supported by SS. So it is unnecessary to support SSR url. Otherwise it would confuse users.

PR #2712, opened 2019-12-03
Improve i18n

studentmain:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature

---

1. TranslateForm function
2. CSV file for easy management
3. Load translation from disk


shadowsocks-csharp/Controller/I18N.cs
```diff
@@ -1,52 +1,113 @@
...
         private static Dictionary<string, string> _strings = new Dictionary<string, string>();
 
-        private static void Init(string res)
+        private static void Init(string res, string locale)
         {
-            using (var sr = new StringReader(res))
+            using (TextFieldParser csvParser = new TextFieldParser(new StringReader(res)))
             {
-                foreach (var line in sr.NonWhiteSpaceLines())
+                csvParser.SetDelimiters(",");
```
celeron533: Data:

| item1 | item2 | item3 |
| aaaaa | bb,cc | d"ef","g,h" |

CSV:
```
item1,item2,item3
aaaaa,"bb,cc","d""ef"",""g,h"""
```

https://tools.ietf.org/html/rfc4180#section-2

```
2.  Definition of the CSV Format

   While there are various specifications and implementations for the
   CSV format (for ex. [4], [5], [6] and [7]), there is no formal
   specification in existence, which allows for a wide variety of
   interpretations of CSV files.  This section documents the format that
   seems to be followed by most implementations:

   1.  Each record is located on a separate line, delimited by a line
       break (CRLF).  For example:

       aaa,bbb,ccc CRLF
       zzz,yyy,xxx CRLF

   2.  The last record in the file may or may not have an ending line
       break.  For example:

       aaa,bbb,ccc CRLF
       zzz,yyy,xxx

   3.  There maybe an optional header line appearing as the first line
       of the file with the same format as normal record lines.  This
       header will contain names corresponding to the fields in the file
       and should contain the same number of fields as the records in
       the rest of the file (the presence or absence of the header line
       should be indicated via the optional "header" parameter of this
       MIME type).  For example:

       field_name,field_name,field_name CRLF
       aaa,bbb,ccc CRLF
       zzz,yyy,xxx CRLF


   4.  Within the header and each record, there may be one or more
       fields, separated by commas.  Each line should contain the same
       number of fields throughout the file.  Spaces are considered part
       of a field and should not be ignored.  The last field in the
       record must not be followed by a comma.  For example:

       aaa,bbb,ccc

   5.  Each field may or may not be enclosed in double quotes (however
       some programs, such as Microsoft Excel, do not use double quotes
       at all).  If fields are not enclosed with double quotes, then
       double quotes may not appear inside the fields.  For example:

       "aaa","bbb","ccc" CRLF
       zzz,yyy,xxx

   6.  Fields containing line breaks (CRLF), double quotes, and commas
       should be enclosed in double-quotes.  For example:

       "aaa","b CRLF
       bb","ccc" CRLF
       zzz,yyy,xxx

   7.  If double-quotes are used to enclose fields, then a double-quote
       appearing inside a field must be escaped by preceding it with
       another double quote.  For example:

       "aaa","b""bb","ccc"
```

studentmain: 图片
图片
图片

Yes, `TextFieldParser` can parse CSV.

celeron533:
I'm not sure if there comes new language, how to do the maintenance?

studentmain:
> how to do the maintenance?
 
Add a new column with language identifier. We can open CSV in Excel or other editor

chenshaoju:
At 4.1.9.0, There is a file name `i18n.csv` automatic generation at the Shadowsocks.exe folder.

I suggest moving to the `ss_win_temp` folder.

studentmain:
Or generate it on demand, add a menu entry, click it to generate i18n.csv, if no external i18n.csv, use builtin one

PR #2212, opened 2019-02-14
Add check box to toggle plugin argument input

studentmain:
## Please follow the guide below

- You will be asked some questions, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *pull request* (like that [x])
- Use *Preview* tab to see how your *pull request* will actually look like

- [x] [Searched](https://github.com/shadowsocks/shadowsocks-windows/search?q=is%3Apr&type=Issues) for similar pull requests
- [x] Compiled the code with Visual Studio

### What is the purpose of your *pull request*?
- [ ] Bug fix
- [x] Improvement
- [ ] New feature

---

### Description of your *pull request* and other information

插件参数默认隐藏以防止用户填错，如图所示
577b39b5-53b5-4dce-ad88-f3e851ffb66d

仅仅是demo，界面什么的还要调整一下

PR #2722, opened 2019-12-10
Implement "Show Plugin Output" checkbox.

rwasef1830:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature

---

This is a simple improvement that allows viewing the plugin output window instead of always hiding it. It can be very useful to debug plugin issues instead of being forced to duplicate the config and run the plugin standalone just to see its output.

### DIFF
--- a/shadowsocks-csharp/Controller/Service/Sip003Plugin.cs
+++ b/shadowsocks-csharp/Controller/Service/Sip003Plugin.cs
@@ -22,7 +22,7 @@ namespace Shadowsocks.Controller.Service
         private bool _started;
         private bool _disposed;
 
-        public static Sip003Plugin CreateIfConfigured(Server server)
+        public static Sip003Plugin CreateIfConfigured(Server server, bool showPluginOutput)
         {
             if (server == null)
             {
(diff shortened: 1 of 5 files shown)


chenshaoju:
Thanks for the send pull request, Please wait for the developer review the code. 😇

studentmain:
Pass all test first

rwasef1830:
@studentmain all tests pass now, my apologies. now it is failing due to unrelated problem with mbedtls test.

studentmain:
Maybe redirect plugin's stdout into log?

Plugin's window can closed by user, which cause plugin exit.

rwasef1830:
Each plugin may have its own way of logging which can be set using plugin arguments. Also no problem if the user closed the plugin window, it will be automatically restarted, so it's not a problem in my opinion.

This is just a simple switch to see the output of the plugin just in case the plugin displayed any messages useful for troubleshooting. It is meant to be used temporarily. And it can accomodate any complicated console output of any plugin.

To make it log to file will complicate things a little (need to log stdout and stderr), stdout or stderr might use non printable characters ... plugin may have its own logging... etc.

What do you think ?

studentmain:
> it's not a problem in my opinion.

Yes, it's not a big problem. The feature is ok (if user know what they did).

Consider what happened in plugin command line inputbox (#2212), I suggest put the switch in right click menu > "Help", just like verbose logging switch.

rwasef1830:
@studentmain The reason why I put the checkbox in that dialog is to allow the user to press Apply and cause restart of the plugin with the restart of the server. If I put it in right-click, and the user choose it, it will not do any effect, and I don't want to cause restart of the server unnecessarily by right-click menu.

This way, all the option that cause restart of the server and plugin are in 1 place (the server config window) and it makes more sense for the user without any surprises. Because the user already expects that the server will be restarted when they press Apply or Ok and we want this to happen when they change the "Show plugin output" setting.

What do you think ?

studentmain:
> If I put it in right-click, and the user choose it, it will not do any effect

Just reload controller in callback function.

>  all the option that cause restart of the server and plugin are in 1 place

This feature is for debugging, so I'd like put it near other debugging option. BTW, server also restart when we toggle system proxy mode......

rwasef1830:
@studentmain Done. It is now as you described to me. Help > Show Plugin Output

studentmain:
Looks good to me, thank you.

celeron533:
A very good way for debugging and tracing error

studentmain:
@celeron533 Let's merge them?

Issue #2729, opened 2019-12-14
Win32Exception when adding plugin on shadowsocks


mathieu001:
### Shadowsocks version / 影梭版本
4.1.8.0

### Environment (Operating system, .NET Framework, etc) / 使用环境（操作系统，.NET Framework等）
Windows 7 Professional(32-bit), .Net Framework 4.7.2

### Steps you have tried / 操作步骤
I am tring to add obfs plugin on shadowsocks and always get a Win32Exception

### What did you expect to see? / 期望的结果
The obfs plugin can be added successfully on my Windows 7 Professional(32-bit)

### What did you see instead? / 实际结果
The obfs plugin can be added on Windows 7 Professional(64-bit), however, it can't be added on Windows 7 Professional(32-bit)

### Config and error log in detail (with all sensitive info masked) / 配置文件和日志文件（请隐去敏感信息）
System.ComponentModel.Win32Exception (0x80004005): The specified executable is not a valid application for this OS platform.

chenshaoju:
If your OS is 32-bit, Please using 32-bit plugin.

PR #2730, opened 2019-12-14
Give a more user friendly message when plugin program file does not e…

celeron533:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature

---

At least gives a hint to user to check whether the plugin program file exists on file system.

### DIFF
--- a/shadowsocks-csharp/Controller/Service/Sip003Plugin.cs
+++ b/shadowsocks-csharp/Controller/Service/Sip003Plugin.cs
@@ -94,18 +94,24 @@ namespace Shadowsocks.Controller.Service
                     return false;
                 }
 
-                if (!File.Exists(_pluginProcess.StartInfo.FileName))
-                {
-                    throw new FileNotFoundException(I18N.GetString("Cannot find the plugin program file"), _pluginProcess.StartInfo.FileName);
-                }
-
                 var localPort = GetNextFreeTcpPort();
                 LocalEndPoint = new IPEndPoint(IPAddress.Loopback, localPort);
 
                 _pluginProcess.StartInfo.Environment["SS_LOCAL_HOST"] = LocalEndPoint.Address.ToString();
                 _pluginProcess.StartInfo.Environment["SS_LOCAL_PORT"] = LocalEndPoint.Port.ToString();
                 _pluginProcess.StartInfo.Arguments = ExpandEnvironmentVariables(_pluginProcess.StartInfo.Arguments, _pluginProcess.StartInfo.EnvironmentVariables);
-                _pluginProcess.Start();
+                try
+                {
+                    _pluginProcess.Start();
+                }
+                catch (System.ComponentModel.Win32Exception ex)
+                {
+                    // do not use File.Exists(...), it can not handle the scenarios when the plugin file is in system environment path.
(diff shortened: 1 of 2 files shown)


shadowsocks-csharp/Data/zh_CN.txt
```diff
@@ -183,6 +183,7 @@ Proxy handshake failed=代理握手失败
 Register hotkey failed=注册快捷键失败
 Cannot parse hotkey: {0}=解析快捷键失败: {0}
 Timeout is invalid, it should not exceed {0}=超时无效，不应超过 {0}
+Cannot find the plugin program file=找不到插件程序文件
```
studentmain: Add program name here?

celeron533: File name will be mentioned in the exception it self in `Sip003Plugin.cs`
```
throw new FileNotFoundException(I18N.GetString("Cannot find the plugin program file"), _pluginProcess.StartInfo.FileName, ex);
```

image

shadowsocks-csharp/Controller/Service/Sip003Plugin.cs
```diff
@@ -94,7 +94,18 @@ public bool StartIfNeeded()
...
                 _pluginProcess.StartInfo.Arguments = ExpandEnvironmentVariables(_pluginProcess.StartInfo.Arguments, _pluginProcess.StartInfo.EnvironmentVariables);
-                _pluginProcess.Start();
+                try
+                {
+                    _pluginProcess.Start();
+                }
+                catch (System.ComponentModel.Win32Exception ex)
+                {
+                    // do not use File.Exists(...), it can not handle the scenarios when the plugin file is in system environment path.
+                    if ((uint)ex.ErrorCode == 0x80004005)   // file not found
```
studentmain: FYI: #2729

celeron533: Thanks, will use native error code to find out the real problem
https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/18d8fbe8-a967-4f1c-ae50-99ca8e491d2d

celeron533: But it is uncommon to see "The specified executable is not a valid application for this OS platform." error. Only when run 64 bit program on 32 bit platform. or just have the bad image.

studentmain: Yes, it's uncommon, but there is😥. Just check if it is FileNotFound, if not, throw error with (or without) native error code so we won't misleading user.

celeron533: Personally prefer to use current implementation to keep everything simple.

studentmain: Copy paste from here
https://docs.microsoft.com/zh-cn/windows/win32/seccrypto/common-hresult-values?redirectedfrom=MSDN


Name | Description | Value
-- | -- | --
S_OK | Operation successful | 0x00000000
E_ABORT | Operation aborted | 0x80004004
E_ACCESSDENIED | General access denied error | 0x80070005
E_FAIL | Unspecified failure | 0x80004005
......|......|.......

Seems we should use native error code...

Issue #2758, opened 2020-01-07
Add abillity to notice breaking change


studentmain:
Prepare for v5, so we can inform user how to migrate to new version at that time.

Possible way to detect breaking change:
- Add &lt;!--__BREAKING_CHANGE__--&gt; in release note
- Add a INSTRUCTION.txt to download

Possible way to inform user:
- Navigate to release page
- Show INSTRUCTION.txt

database64128:
Closed in #2991

PR #2778, opened 2020-01-18
Check program is updated

studentmain:

- [ ] Bug fix
- [ ] Improvement
- [x] New feature

---

Part of #2758, we can detect we're updated now.

图片


celeron533:
当时在gui-config文件中留下版本号就是为了检查用户从哪个版本更新过来的，万一版本间有较大的配置变更可以做迁移。还有个好处就是当用户贴config到github issue也能比较方便的看到版本。

shadowsocks-csharp/Model/Configuration.cs
```diff
@@ -78,6 +82,10 @@ public static Configuration Load()
                 string configContent = File.ReadAllText(CONFIG_FILE);
                 Configuration config = JsonConvert.DeserializeObject<Configuration>(configContent);
                 config.isDefault = false;
+                if (UpdateChecker.Asset.CompareVersion(UpdateChecker.Version, config.version) > 0)
+                {
+                    config.updated = true;
```
celeron533: How about launch an event such as "OnUpdated" rather than a flag?
So we can register events such as: 1. show balloon. 2. run migration script when necessary 3. show change log if necessary

studentmain: Good.

shadowsocks-csharp/Model/Configuration.cs
```diff
@@ -78,6 +82,10 @@ public static Configuration Load()
                 string configContent = File.ReadAllText(CONFIG_FILE);
                 Configuration config = JsonConvert.DeserializeObject<Configuration>(configContent);
                 config.isDefault = false;
+                if (UpdateChecker.Asset.CompareVersion(UpdateChecker.Version, config.version) > 0)
```
celeron533: If user is updated from a legacy version which does not contains 'version' element in `gui-config` file. The code should still work.

shadowsocks-csharp/View/MenuViewController.cs
```diff
@@ -604,14 +609,6 @@ private void CheckUpdateForFirstRun()
             updateChecker.CheckUpdate(config, 3000);
         }
 
-        private void ShowFirstTimeBalloon()
-        {
-            _notifyIcon.BalloonTipTitle = I18N.GetString("Shadowsocks is here");
-            _notifyIcon.BalloonTipText = I18N.GetString("You can turn on/off Shadowsocks in the context menu");
```
celeron533: Remove the translation line in csv file if necessary

celeron533:
@studentmain you could merge this pull request if you are confident.

shadowsocks-csharp/Model/Configuration.cs
```diff
@@ -37,9 +37,14 @@ public class Configuration
         public HotkeyConfig hotkey;
 
         private static readonly string CONFIG_FILE = "gui-config.json";
+
+        [JsonIgnore]
+        public bool updated = false;
```
celeron533: Should it be moved to UpdateChecker.cs ?
Reason:
1. This flag is set by comparing the local version and remote version.
2. It is only checked when checking updates.
3. No necessary for persistence and the flag is not widely used in SS.

studentmain: 1. Binary version and config version, both are local.
2. Then manually download and replace won't trigger it.
3. Agree.

I'd like refactor UpdateChecker(maybe Configuration too?) to single instance pattern or static class before move it...

studentmain: I'll keep current behavior, refactor will in another PR.

PR #2783, opened 2020-01-27
Using NLog

celeron533:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature

---

Replace the current logger to NLog with nlog.config.
Now just using legacy `Verbose log` switch. In future it could be more detailed log levels.
Main executable file size is about 2.66 MB.


chenshaoju:
The quick test is passed.
But there is a `NLog.config` file with Shadowsocks.exe same folder, It's correct?

celeron533:
> But there is a NLog.config file with Shadowsocks.exe same folder, It's correct?

Yes, will be extracted when it does not exist.
Then SS calls NLog.LogManager to reload this config file to activate.
Next time, when start SS, LogManager will load this file as its nature behavior.

studentmain:
Consider move config file into temporary directory and/or always extract it.

celeron533:
> Consider move config file into temporary directory

Form NLog official document:
```
For a stand-alone *.exe application, files are searched as follows:

1. standard application configuration file (usually applicationname.exe.config)
2. applicationname.exe.nlog in application’s directory
3. NLog.config in application’s directory (Name sensitive; using docker dotnet core)
4. NLog.dll.nlog in a directory where NLog.dll is located (only if NLog isn't installed in the GAC)
```
Could also explicitly load from specific file.
`NLog.LogManager.Configuration = new NLog.Config.XmlLoggingConfiguration("nlog.config");`

However, it is not encouraged "always extract" the NLog.config file because advanced user may added their own targets or rules with log level in the file. (Unless we have break change that require rewrite the NLog config.)

For putting things in temp folder... Could not find a strong reason to argue why or why not do this.

studentmain:
Question: Should we allow advanced user edit config:
How if user broke the config file? How if user removed LogForm target by accident?

If we allow user edit config, then we can keep current behavior.

shadowsocks-csharp/Data/NLog.config
```diff
@@ -0,0 +1,12 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<!-- configuration may reset after shadowsocks upgrade -->
```
celeron533: @studentmain here is the NLog config file. Allow user to modify some of the part but not encouraged. Once the code is merged, corresponding Wiki page would be created.

studentmain: Ok, if user broke the config, we can say "We warned you". It won't be problem now.

PR #2819, opened 2020-03-01
show message when cipher not exist

studentmain:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

---

When a cipher not exist (user migrate from a too old version), show a warning message instead of crash.

### DIFF
--- a/shadowsocks-csharp/View/ConfigForm.cs
+++ b/shadowsocks-csharp/View/ConfigForm.cs
@@ -79,7 +79,14 @@ namespace Shadowsocks.View
             public static EncryptionMethod GetMethod(string name)
             {
                 if (!init) Init();
-                return methodByName[name];
+                bool success = methodByName.TryGetValue(name, out EncryptionMethod method);
+                if (!success)
+                {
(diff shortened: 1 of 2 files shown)


shadowsocks-csharp/View/ConfigForm.cs
```diff
@@ -79,7 +79,14 @@ private static void Init()
             public static EncryptionMethod GetMethod(string name)
             {
                 if (!init) Init();
-                return methodByName[name];
+                bool ok = methodByName.TryGetValue(name, out EncryptionMethod method);
```
celeron533: bool success = ...

studentmain: Done.

By the way, seems NLog cause some strange deadlock in ShadowsocksController.Reload(). Open ConfigForm, click OK will trigger it.

celeron533: There was a log output file issue when opening config form. Fixed in recent commits.

Issue #2841, opened 2020-04-01
4.1.9.3 autostart issue


vanyaindigo:
The app doesn't start with the system while autostart is switch on.

chenshaoju:
Please check registry `\HKCU\Software\Microsoft\Windows\CurrentVersion\Run`,There is a `Shadowsocks_xxxxxxxx` entry?

vanyaindigo:
Yes, there is such entry

chenshaoju:
I tested on my Windows 10 1909 (18363.720), But it can't be repeated.

It's your computer install some computer manager software, Like 360 Security Center or Tencent PC Manager?

vanyaindigo:
Nothing installed and with ver. 4.1.9.2 all was fine.

chenshaoju:
Odd, Please quit Shadowsocks,Open registry and delete `Shadowsocks_xxxxxx` from `\HKCU\Software\Microsoft\Windows\CurrentVersion\Run`, And start Shadowsocks again and set it to auto start again?

vanyaindigo:
Still not starting

chenshaoju:
This should not happen. 

If you manually create an entry name something else like `test` with your Shadowsocks path, It's can work?

example:
![image](https://user-images.githubusercontent.com/515763/78229745-f50ee580-7502-11ea-9814-d727f62a2c04.png)

vanyaindigo:
I think this is somehow refer with new localization

chenshaoju:
What is your operating system language or location settings?

vanyaindigo:
Russian

studentmain:
> Russian, is it possible to switch back to english?

Help-> Write translate template, then delete Russian column in generated file.

studentmain:
Is there anything in logfile?

vanyaindigo:
> Is there anything in logfile?

Nothing

celeron533:
tag as `bug` for tracking.
It is possible that this new binary release is not trusted by the anti-virus and does not allowed to be added in startup?

vanyaindigo:
There is an error in system log:
```
Контейнер ошибки , тип 0
Имя события: CLR20r3
Отклик: Нет данных
Идентификатор CAB: 0

Сигнатура проблемы:
P1: Shadowsocks.exe
P2: 4.1.9.3
P3: 5e835a75
P4: mscorlib
P5: 4.8.4150.0
P6: 5e176d21
P7: 1693
[...]

PR #2846, opened 2020-04-05
NLogConfig - TouchAndApplyNLogConfig with Exception handler

snakefoot:

- [x] Bug fix

---

### Description

Avoid application crash when not having file permissions to setup default NLog.config. See also #2841

### DIFF
--- a/shadowsocks-csharp/Model/NlogConfig.cs
+++ b/shadowsocks-csharp/Model/NlogConfig.cs
@@ -108,11 +108,20 @@ namespace Shadowsocks.Model
         /// </summary>
         public static void TouchAndApplyNLogConfig()
         {
-            if (!File.Exists(NLOG_CONFIG_FILE_NAME))
+            try
             {
+                if (File.Exists(NLOG_CONFIG_FILE_NAME))
+                    return; // NLog.config exists, and has already been loaded
+
                 File.WriteAllText(NLOG_CONFIG_FILE_NAME, Properties.Resources.NLog_config);
-                LogManager.LoadConfiguration(NLOG_CONFIG_FILE_NAME);
             }
+            catch (Exception ex)
+            {
+                NLog.Common.InternalLogger.Error(ex, "[shadowsocks] Failed to setup default NLog.config: {0}", NLOG_CONFIG_FILE_NAME);
+                return;
+            }
+
+            LoadConfiguration();    // Load the new config-file
         }
 
         /// <summary>


celeron533:
Thanks @snakefoot , is has been fixed in 5bc4a0b1cc595b83f81149bf656de48f93758296

shadowsocks-csharp/Model/NlogConfig.cs
```diff
@@ -108,11 +108,20 @@ private static XmlNode SelectSingleNode(XmlDocument doc, string xpath)
...
                 File.WriteAllText(NLOG_CONFIG_FILE_NAME, Properties.Resources.NLog_config);
-                LogManager.LoadConfiguration(NLOG_CONFIG_FILE_NAME);
             }
+            catch (Exception ex)
+            {
+                NLog.Common.InternalLogger.Error(ex, "[shadowsocks] Failed to setup default NLog.config: {0}", NLOG_CONFIG_FILE_NAME);
+                return;
+            }
+
+            LoadConfiguration();    // Load the new config-file
```
celeron533, line 124: There may not have the NLog config file due to the previous write permission issue.

snakefoot:
It only performs the LoadConfiguration  if it was successful in creating one.


Sent from my Sony Xperia


---- Allen Zhu wrote ----


@celeron533 commented on this pull request.

________________________________

In shadowsocks-csharp/Model/NlogConfig.cs<https://github.com/shadowsocks/shadowsocks-windows/pull/2846#discussion_r403913282>:

>              }
+            catch (Exception ex)
+            {
+                NLog.Common.InternalLogger.Error(ex, "[shadowsocks] Failed to setup default NLog.config: {0}", NLOG_CONFIG_FILE_NAME);
+                return;
+            }
+
+            LoadConfiguration();    // Load the new config-file


There may not have the NLog config file due to the previous write permission issue.

shadowsocks-csharp/Model/NlogConfig.cs
snakefoot: It only performs the `LoadConfiguration`  if it was successful in creating one. So if it can write to the location, then it should also be able to read.

studentmain: So the routine is like this? I think this change is ok then.

When write error: Line 113->116 throw error ->118 catch error->120->121 return
When config exist: (config auto loaded by NLog itself at program start) 113->114 return
When config not exist: 113->116->124 load config->return

snakefoot: Yes it returns without doing anything for all cases. Except for the very first time where it successfully have created a new default `NLog.config`.

celeron533: Since the 5bc4a0b1cc595b83f81149bf656de48f93758296 is committed to fix the CurrentDirectory issue. The NLog config file is always create/read/write on the same place in file system. Will merge the code soon.

PS:
Before the previous autostart issue fix, SS may create the NLog config file in the wrong place, such as `System` folder and request the NLog to load it.
Then the following code change the CurrentDirectory to `Application.StartPath`
https://github.com/shadowsocks/shadowsocks-windows/blob/5bc4a0b1cc595b83f81149bf656de48f93758296/shadowsocks-csharp/Program.cs#L86
Which made the NLog file location inconsistent. When user change the log level and NLog need to reload the config file, may fail.

shadowsocks-csharp/Model/NlogConfig.cs
```diff
@@ -108,11 +108,20 @@ private static XmlNode SelectSingleNode(XmlDocument doc, string xpath)
         /// </summary>
         public static void TouchAndApplyNLogConfig()
         {
-            if (!File.Exists(NLOG_CONFIG_FILE_NAME))
+            try
             {
+                if (File.Exists(NLOG_CONFIG_FILE_NAME))
+                    return; // NLog.config exists, and has already been loaded
```
celeron533, line 114: File exists, but it doesn't mean it has been loaded by the program.

PR #2855, opened 2020-04-18
New option for ss:// URL association

database64128:

- [ ] Bug fix
- [ ] Improvement
- [x] New feature

---

Added an option to allow association with ss:// URLs.

This PR was made at the request of @studentmain


chenshaoju:
The quick test is pass, Base appveyor version 4.1.10.4.

But there is a small issue: when ss:// URI is invalid, There is no error message.

shadowsocks-csharp/Controller/System/ProtocolHandler.cs
```diff
@@ -0,0 +1,104 @@
...
+        // see https://stackoverflow.com/questions/12945805/odd-c-sharp-path-issue
+        private static readonly string ExecutablePath = Assembly.GetEntryAssembly().Location;
+
+        // TODO: Elevate when necessary
+        public static bool Set(bool enabled)
+        {
+            RegistryKey ssURLAssociation = null;
+            try
+            {
+                ssURLAssociation = Registry.CurrentUser.CreateSubKey(@"SOFTWARE\Classes\ss", RegistryKeyPermissionCheck.ReadWriteSubTree);
```
celeron533, line 27: Perhaps using full name "Shadowsocks" is better

shadowsocks-csharp/Controller/System/ProtocolHandler.cs
```diff
@@ -0,0 +1,104 @@
...
+                if (ssURLAssociation == null)
+                {
+                    logger.Error(@"Failed to create HKCU\SOFTWARE\Classes\ss");
```
celeron533, line 30: More friendly message "Failed to create HKCU\SOFTWARE\Classes\Shadowssocks to register the URL association"

shadowsocks-csharp/Controller/System/ProtocolHandler.cs
```diff
@@ -0,0 +1,104 @@
...
+                }
+            }
+        }
+
+        public static bool Check()
+        {
+            RegistryKey ssURLAssociation = null;
+            try
+            {
+                ssURLAssociation = Registry.CurrentUser.OpenSubKey(@"SOFTWARE\Classes\ss", true);
```
celeron533, line 73: could extract the registry path into a constant string

database64128:
> The quick test is pass, Base appveyor version 4.1.10.4.
> 
> But there is a small issue: when ss:// URI is invalid, There is no error message.

Actually the original "Import URL from clipboard" also shows no error message. I added messages for both in a1e4245.

shadowsocks-csharp/Controller/System/ProtocolHandler.cs
database64128: > Perhaps using full name "Shadowsocks" is better

By using "Shadowsocks" you are registering Shadowsocks:// instead of ss://.

shadowsocks-csharp/Controller/System/ProtocolHandler.cs
database64128: > More friendly message "Failed to create HKCU\SOFTWARE\Classes\Shadowssocks to register the URL association"

I'll make a quick change to it. But just so you know, this operation shouldn't normally fail.

shadowsocks-csharp/Controller/System/ProtocolHandler.cs
celeron533: Thanks @database64128 , I made a mistake...

Note:
Outline or other software may also register the protocol
image

shadowsocks-csharp/Controller/System/ProtocolHandler.cs
database64128: Changes made in 3652e7f.

shadowsocks-csharp/Controller/System/ProtocolHandler.cs
database64128: > Outline or other software may also register the protocol

Just tested with Outline installed. The current behavior of our implementation is overwriting existing association, which is not perfect, but good enough. As far as I know, there is no way to associate a URI scheme with more than one Win32 application. The [API for that](https://docs.microsoft.com/en-us/windows/uwp/launch-resume/handle-uri-activation) is only available for UWP apps.

database64128: Also note that Outline overwrites the URI association on boot. Apparently they are using the same implementation, with a more aggressive approach and no way to turn off.

database64128: Our implementation checks if ss:// is associated with itself on boot to determine the status of association, and shows it with `MenuItem.Checked`. It's up to the user to associate the URI scheme or not.

studentmain: > with a more aggressive approach and no way to turn off

We need open an issue for it.

celeron533:
Name `pipe` or `pipeserver` may a bit confusion with existing `StartPipe`, `PipeConnection...` network related communications.

database64128:
> Name `pipe` or `pipeserver` may a bit confusion with existing `StartPipe`, `PipeConnection...` network related communications.

Resolved in 4bd33db.

database64128:
Any plan to merge? I've been using the build on my PCs and everything seems good.

celeron533:
Merged. Thank you very much @database64128 !

PR #2982, opened 2020-10-11
📃 Exclude @cn from PAC proxied list

### LINKED ISSUE #2971: Geosite包含大量国内地址，未来有改进计划吗？
### 简要描述问题
新版采用的geosite domain list中包含大量国内地址，自动生成pac时会被包含在规则内，导致不必要的代理。
例如，所有暴雪/网易游戏相关的地址。目前可以用user-rule把所有相关的地址都单独拎出来设成例外，但是geosite里面包含的所有其他大陆地址对于国内用户来说也太不方便了吧，有什么改进计划吗？

### 环境

- Shadowsocks客户端版本：
4.2.0.1
- 操作系统版本：
Windows 10
- .NET版本：
4.8

### 操作步骤
1. 升级SS至4.2.0.1
2. 删除并重生成pac
3. 打开魔兽世界客户端wow.exe并尝试登陆

### 期望的结果
快速登陆至人物界面

### 实际结果
卡在“正在连接”很久，然后登陆至人物界面

### 配置文件和日志文件（请隐去敏感信息）
无

### 临时解决办法
在user-rule.txt中将所有相关的地址全部标记为例外。
```
@@||blizzard.cn
@@||blzstatic.cn
@@||battlenet.com.cn
@@||bnet.163.com
@@||wowchina.com
@@|http://wow.nosdn.127.net
@@|https://wow.nosdn.127.net
@@|http://blz.nosdn.127.net
@@|https://blz.nosdn.127.net
@@|http://di.res.netease.com
@@|https://di.res.netease.com
@@|http://heroes.nos.netease.com
@@|https://heroes.nos.netease.com
@@|http://diablo3.nosdn.127.net
@@|https://diablo3.nosdn.127.net
@@|http://hearthstone.nosdn.127.net
@@|https://hearthstone.nosdn.127.net
@@|http://overwatch.nosdn.127.net
@@|https://overwatch.nosdn.127.net
@@|http://sc2.nosdn.127.net
@@|https://sc2.nosdn.127.net
```

### LINKED ISSUE #2974: 新版本使用geosite导致winstore无法加载
### 简要描述问题

更新版本后使用新版的geosite数据，在pac模式下无法正常打开winstore，并且大略的查询几个常见的地址都并非被墙网站。
希望能换回之前的gfwlist或者提供选项。

### 环境

- Shadowsocks客户端版本：4.2.0.1
- 操作系统版本：Win10 2004
- .NET版本：

### 操作步骤

1. 更新ss-win + 使用geosite更新本地pac
1. 设置为pac模式
1. 打开winstore

### 期望的结果

正常打开

### 实际结果

疑似因为通过代理请求导致winstore无法正常加载

### 配置文件和日志文件（请隐去敏感信息）

```
在此粘贴日志
```

### LINKED ISSUE #2981: Why did we change PAC source to `geosite`?
### Describe the bug
The new PAC source proxy some domains that can be accessed without shadowsocks in China, probably because the new source's PAC list isn't as complete as the old one.

The domains that are mistakenly proxyed I found:
1. Microsoft official website.
2. Bizzard Battle.net's downloading source. (when I turned off Shadowsocks, the game download speed raised to 3MB/s, whereas only 30KB/s with the PAC mode opened.)

### Environment

- Shadowsocks client version: 4.2.0.1
- OS version: windows 10
- .NET version:

database64128:

- [ ] Bug fix
- [x] Improvement
- [ ] New feature

---

- PAC proxied list now only contains domain names from geolocation-!cn without a `cn` attribute
- Closes #2971.
- Closes #2974.
- Closes #2981.

### DIFF
--- a/shadowsocks-csharp/Controller/Service/GeositeUpdater.cs
+++ b/shadowsocks-csharp/Controller/Service/GeositeUpdater.cs
@@ -39,15 +39,18 @@ namespace Shadowsocks.Controller
         private static HttpClient httpClient;
         private static readonly string GEOSITE_URL = "https://github.com/v2fly/domain-list-community/raw/release/dlc.dat";
         private static readonly string GEOSITE_SHA256SUM_URL = "https://github.com/v2fly/domain-list-community/raw/release/dlc.dat.sha256sum";
+        private static readonly DomainObject.Types.Attribute geositeExcludeAttribute;
         private static byte[] geositeDB;
 
         public static readonly Dictionary<string, IList<DomainObject>> Geosites = new Dictionary<string, IList<DomainObject>>();
 
         static GeositeUpdater()
         {
-            //socketsHttpHandler = new SocketsHttpHandler();
-            //httpClient = new HttpClient(socketsHttpHandler);
-            
+            geositeExcludeAttribute = new DomainObject.Types.Attribute
+            {
+                Key = "cn",
+                BoolValue = true
+            };
             if (File.Exists(DATABASE_PATH) && new FileInfo(DATABASE_PATH).Length > 0)
             {
                 geositeDB = File.ReadAllBytes(DATABASE_PATH);
@@ -241,6 +244,9 @@ var __RULES__ = {JsonConvert.SerializeObject(gfwLines, Formatting.Indented)};
             List<string> ret = new List<string>(domains.Count + 100);// 100 overhead
             foreach (var d in domains)
             {
+                if (d.Attribute.Contains(geositeExcludeAttribute))
+                    continue;
+
                 string domain = d.Value;
 
                 switch (d.Type)


chenshaoju (approved):
The quick test is passed, base `Shadowsocks-4.2.0.40-Release.exe`, the domain like `blizzard.cn`  etc, has not through the proxy.

But some domains like `qualcomm.cn`, `dynacw.com.cn`, `verisign.com.cn`, `microsoft.com` etc, still in the list.

I'm curious about these domains.

Anyway, :+1: Looks good to me, but someone else must approve.

database64128:
> But some domains like `qualcomm.cn`, `dynacw.com.cn`, `verisign.com.cn`, `microsoft.com` etc, still in the list.

The domains you mentioned don't have the `cn` attribute. So it's the expected behavior that they are included in the proxied list. And they actually resolves to non-CN IP addresses. So that's not a problem at all.

chenshaoju:
image

Odd... But anyway this is Geosite issue. 😅

Stzx (approved):
In fact, for the problems in the third party, I personally think it should be solved only upstream. But when the upstream processing has not yet resulted, I think we can apply some temporary fixes.

studentmain (approved):

database64128:
> In fact, for the problems in the third party, I personally think it should be solved only upstream. But when the upstream processing has not yet resulted, I think we can apply some temporary fixes.

See https://github.com/v2fly/domain-list-community/issues/235#issuecomment-700227629.

In a nutshell, `geolocation-!cn` is just a list of non-CN entities' domain names. It is supposed to contain non-CN entities' Chinese domain names. The `cn` attribute is the upstream's answer to our use case, where we only want non-CN domain names in the list, and don't want Chinese domain names.

ziruizhuang:
I think the problem is that the original motivation of this application is to accelerate an endpoint's Internet connection to specific sites through a preconfigured faster channel. And in normal circumstances, the quantity of sites with extremely slow connection speed that must rely on a faster channel is relatively smaller compared to a vast range of sites that can be accessed with reasonable speed. So, for this basic demand, an access-list based PAC rule set works pretty well. Even if there are conditions where the access list misses a site, a user can always switch to Global mode, and the effort is trivial. 

I understand, users may need more advanced features, such as accelerate all remote traffic through a faster channel, so that the general quality of experience is better. However, I question the use of domain-based methods to achieve this goal.

And in my opinion, domain-based methods are not well suited. This is because that domain names changes a lot. Even if a company holds a well-known long-standing domain, they can always choose to deploy new servers at proximity or purchase new CDN services. This is exactly why domains are invented in the first place. And in situation like this, I highly doubt that a community-based voluntary work like GeoSite can keep its pace with the swift change of the Internet players. 

This causes trouble for some major services with CDN support, for example, Microsoft Store, Windows Update, Windows Feedback, Windows Mail, etc. Note that even after I have already whitelisted a bunch of Microsoft and Windows related domains, the Windows Update is still affected. So, the effort to find out which domain these services use and to whitelist them is not trivial, especially for general end-users.

A more realistic method would be the use of Geo-Location IP profiles. The regions of Autonomous Systems are highly correlated to geolocations, and the interconnection structures are much much more stable than domains. 

If there is no practical implementation in the near future, it is better providing an access-list based method as default and give users a choice to opt for the experimental Geo-based method.

database64128:
> However, I question the use of domain-based methods to achieve this goal.

Domain-based methods are all our infrastructure can support right now.

> This causes trouble for some major services with CDN support

This PR addresses that issue by excluding the `cn` attribute, which marks domain names that have Chinese CDNs.

> it is better providing an access-list based method as default and give users a choice to opt for the experimental Geo-based method.

`gfwlist` has been poorly maintained for a long time. And in my personal experience, I ended up maintaining a long list of user rules. With `geosite`, the problem is mostly mitigated, because most commonly known sites are included. And anyone can contribute and they process PRs fast.

> A more realistic method would be the use of Geo-Location IP profiles.

Once again, anyone is more than welcome to work with us to add routing.

ziruizhuang:
> Domain-based methods are all our infrastructure can support right now.

That is fine. I understand that new features need time and work before it comes to a stable version. And I do appreciate the hard work done by the developers.

> This PR addresses that issue by excluding the `cn` attribute, which marks domain names that have Chinese CDNs.

Unfortunately, this PR **DID NOT** address that issue. The 'cn' attribute is incomplete, even for the domains of some well-known major service providers like Microsoft. And as I have pointed out, I personally already whitelisted a bunch of MS domains, yet still to find there are more to be whitelisted. For more general users, identifying affected domains and whitelisting them needs certainly **NOT** trivial effort.

> > it is better providing an access-list based method as default and give users a choice to opt for the experimental Geo-based method.
> 
> `gfwlist` has been poorly maintained for a long time. And in my personal experience, I ended up maintaining a long list of user rules. With `geosite`, the problem is mostly mitigated, because most commonly known sites are included. And anyone can contribute and they process PRs fast.

As pointed out previously, domain-based Geo rules are born with these problems. No matter how fast they process PR request, still people will find their services are broken or slowed down, as services providers can and will change their domains or associated IPs any time they want, without notifying GeoSite. This will break general users' ability accessing the services they need.

In my personal experience, I ended up maintaining a list of user-rules whitelisting service domains with CDN support. 
Moreover, to find out some websites that must go through faster channels are relatively easier than isolating domains used by services highly integrated into operating systems.

> > A more realistic method would be the use of Geo-Location IP profiles.
> 
> Once again, anyone is more than welcome to work with us to add routing.

And once again, 
>>If there is no practical implementation in the near future, it is better providing an access-list based method as default and give users a choice to opt for the experimental Geo-based method.

Since the code is already there, it seems like a no-brainer to support the switching between GeoSite and access-rule based methods. I am not saying we should going back to a decade ago. Instead, I am saying that we should give users the choice to switch between these methods before the new feature comes to its perfection.

database64128:
> The 'cn' attribute is incomplete, even for the domains of some well-known major service providers like Microsoft.

It's actually quite complete for most major service providers.

`gfwlist` is not coming back, period.

ziruizhuang:
Still, I am not saying we should go back to previous methods. I am saying they could be co-existing for some more time, to let developers improve GeoSite-based methods.

I am trying to isolate which domain is affected during Windows Update. Yet after turning the proxy off, the Windows Update progress proceeds so fast that I haven't got the chance to isolate it. Hopefully, next month I will get the chance.

ziruizhuang:
For example, in #2987 there is a domain with CDN support: 'dl.delivery.mp.microsoft.com'
> The domain name dl.delivery.mp.microsoft.com has the cn attribute and is therefore excluded from the PAC since v4.2.1.0. I checked my local pac.txt and it's not in it.
Although dl.delivery.mp.microsoft.com has the cn attribute, and it is excluded from the PAC, the root domain 'microsoft.com' is still inside PAC file, and it will match 'dl.delivery.mp.microsoft.com' upon request.

ziruizhuang:
The root cause is that, in my opinion, the PAC rules are black-list based logic, while the composition of GeoSite list is kind of white-list based logic. So, the transition from previous access-rule based method to this GeoSite-based method should be more cautious. 

Again, using GeoSite has good motivation, the devotion and sacrifice from developers are appreciated, but this has nothing to do with the fact that GeoSite-based methods still needs more time and work to reach its perfection.

PR #2991, opened 2020-10-17
🎂 ForwardProxyView + HotkeysView + OnlineConfigView + VersionUpdatePromptView

database64128:

- [x] Bug fix
- [x] Improvement
- [x] New feature

---

- Infrastructure: use one `HttpClient` instance throughout the lifecycle
- Version update: rewrite the update service and add update prompt window
- Server sharing: add copy link button
- Dependencies: add `ReactiveUI.Events.WPF`, `ReactiveUI.Fody`, `ReactiveUI.Validation`, `WPFLocalizeExtension`, `MdXaml`


chenshaoju (approved):
The quick test is pass, base appveyor `Shadowsocks-4.2.0.78.zip` version.

👍  Looks good to me, but someone else must approve.

shadowsocks-csharp/Controller/Service/UpdateChecker.cs
```diff
@@ -1,254 +1,174 @@
...
 using Shadowsocks.Model;
 using Shadowsocks.Util;
+using Shadowsocks.Views;
 
 namespace Shadowsocks.Controller
 {
     public class UpdateChecker
     {
-        private static Logger logger = LogManager.GetCurrentClassLogger();
+        private readonly Logger logger;
```
Stzx, line 21: Why convert to member variables?

database64128: Because it should've been a read-only private field in the first place. The `UpdateChecker` class has only one instance managed by `MenuViewController` throughout the lifecycle. There's no need to have any static fields here. Having unnecessary static fields breaks the code hierarchy.

Stzx (approved):

chenshaoju:
After 24 hours test, There looks have an issue:

Everything when I start Shadowsocks, the "No available update" message bubble always popup.

If I disable `Auto check update` then no more message bubble when startup.

IMHO, If not user manually check update then popup "No available update" is no necessity.

database64128:
> IMHO, If not user manually check update then popup "No available update" is no necessity.

@chenshaoju Thanks for the feedback! I'm adding a check to prevent the no available updates pop-up from showing up when it
s an automatic check.

PR #3012, opened 2020-11-01
remove duplicate startup entries

### LINKED ISSUE #3011: Duplicate startup entry on registry
### Describe the bug

after upgrading to v4.3, I'm seeing `shadowsocks is already running...` error on my windows startup. I've checked `HKEY_CURRENT_USER\SOFTWARE\Microsoft\Windows\CurrentVersion\Run` and there's two shadowsocks entry on the registry

### Environment

- Shadowsocks client version: v4.3.1.0
- OS version: Windows 10

### What did you expect to see?

you should check whether there's an entry containing the path of current executable on registry or not when adding a new entry (or removing it)

### What did you see instead?

two shadowsocks on `HKEY_CURRENT_USER\SOFTWARE\Microsoft\Windows\CurrentVersion\Run`

ahmadalli:

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

---

fixes #3011

### DIFF
--- a/shadowsocks-csharp/Controller/System/AutoStartup.cs
+++ b/shadowsocks-csharp/Controller/System/AutoStartup.cs
@@ -74,14 +74,14 @@ namespace Shadowsocks.Controller
                 var check = false;
                 foreach (var valueName in runKey.GetValueNames())
                 {
-                    if (string.Equals(valueName, Key, StringComparison.InvariantCultureIgnoreCase))
+                    if (valueName.Equals(Key, StringComparison.InvariantCultureIgnoreCase))
                     {
                         check = true;
                         continue;
                     }
 
                     // Remove other startup keys with the same executable path. fixes #3011 and also assures compatibility with older versions
-                    if (string.Equals(runKey.GetValue(valueName).ToString(), Program.ExecutablePath, StringComparison.InvariantCultureIgnoreCase))
+                    if (Program.ExecutablePath.Equals(runKey.GetValue(valueName).ToString(), StringComparison.InvariantCultureIgnoreCase))
                     {
                         runKey.DeleteValue(valueName);
                         runKey.SetValue(Key, Program.ExecutablePath);
--- a/shadowsocks-csharp/Controller/System/AutoStartup.cs
+++ b/shadowsocks-csharp/Controller/System/AutoStartup.cs
@@ -14,7 +14,7 @@ namespace Shadowsocks.Controller
 
         // Don't use Application.ExecutablePath
         // see https://stackoverflow.com/questions/12945805/odd-c-sharp-path-issue
-        
+
         private static string Key = "Shadowsocks_" + Program.ExecutablePath.GetHashCode();
 
         public static bool Set(bool enabled)
@@ -27,6 +27,17 @@ namespace Shadowsocks.Controller
                 {
                     logger.Error(@"Cannot find HKCU\Software\Microsoft\Windows\CurrentVersion\Run");
                     return false;
+                }
+                // Remove other startup keys with the same executable path. fixes #3011
+                foreach (var valueName in runKey.GetValueNames())
+                {
+                    if (valueName == Key)
+                        continue;
+
+                    if (string.Equals(runKey.GetValue(valueName).ToString(), Program.ExecutablePath, StringComparison.InvariantCultureIgnoreCase))
+                    {
+                        runKey.DeleteValue(valueName);
+                    }
                 }
                 if (enabled)
                 {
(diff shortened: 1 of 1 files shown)


chenshaoju (approved):
:+1: Looks good to me, but someone else must approve.

database64128:
The status of the `AutoStartupItem` checkbox is checked on app startup.

https://github.com/shadowsocks/shadowsocks-windows/blob/5f760a5e5740862bbef6f1ca1e393dbc9ce3f5b8/shadowsocks-csharp/View/MenuViewController.cs#L358

It's impossible to have duplicate registry entries, unless you manually add them after starting the app. Or the check process may have failed to cover some edge cases.

ahmadalli:
as I've explained on #3011, I've experienced it after upgrade. I'm not sure from which version but I guess it was from 4.2.1 to 4.3.0

ahmadalli:
also there's an issue with the `Check` method code:

https://github.com/shadowsocks/shadowsocks-windows/blob/73fa9205ed6973e40b6372d6053fa0780323ee55/shadowsocks-csharp/Controller/System/AutoStartup.cs#L75-L88

if the loop reaches the `key` first, it won't go further and won't reach the `Shadowsocks` key therefore it won't delete it

database64128:
Please take another look at `AutoStartup.cs`. `AutoStartup.Check()` does almost exactly what you added in your PR, except it doesn't check for or delete duplicate entries.

As for the issue, I doubt if it's caused by the update. Can you reproduce it and provide detailed steps?

ahmadalli:
the `Check` method is the right place to do what I did in this PR. but it has a corner case.

lets assume `runList` array is this: `["key1", "key2", ..., "shadowsocks_xyz <this is the Key variable", "key3", ..., "shadowsocks"]`

whe the foreach loop reaches the `shadowsocks_xyz`, it'll return `true` and won't reach the `shadowsocks` key and won't get a chance to remove the key.

database64128:
> if the loop reaches the `key` first, it won't go further and won't reach the `Shadowsocks` key therefore it won't delete it

It's impossible to have duplicate registry entries in the first place, unless you manually add them after starting the app, which shouldn't be covered by the checker.

ahmadalli:
> It's impossible to have duplicate registry entries in the first place, unless you manually add them after starting the app, which shouldn't be covered by the checker.

the case `check` is covering is exactly fixing duplicate startup entries (different names, same path) which is caused by upgrading

ahmadalli:
I've fixed the issue of check method on my latest commit. the only difference is that I don't just check for `Shadowsocks` on the registry, I check all the entries to see if they have the same path as current executable

ahmadalli:
here's a detailed way to reproduce this issue:

1. install v4.1.9.2 (any version before commit ced49e2de12b70908c3ce43a33cb4130fa177879) on `C:/Temp`
1. open shadowsocks and click on `Start on Boot` to enable it
1. close shadowsocks, remove the contents of `C:/Temp` directory
1. install `4.2.0.1` (this is the version I used for testing. later versions should be fine)
1. open shadowsocks
1. `Start on Boot` is inactive
1. check the contents of `HKEY_CURRENT_USER\SOFTWARE\Microsoft\Windows\CurrentVersion\Run`
1. click on `Start on Boot` to enable it
1. check the contents of `HKEY_CURRENT_USER\SOFTWARE\Microsoft\Windows\CurrentVersion\Run` again

and there will be two `Shadowsocks_...` entries on the registry with the same path

ahmadalli:
and the current `Check` method doesn't remove the other `shdowsocks_...` key because it

1. it gives up checking after it finds the `Key` variable on the registry
1. there's no check of other `Shadowsocks_...` entries. just checks if there's any `Shadowsocks` entry (which there isn't in this case)

database64128:
Thank you for your contribution and detailed explanations! 😊

ahmadalli:
yw ;)

if you're going to merge it now, just let me adjust one more thing

shadowsocks-csharp/Controller/System/AutoStartup.cs
```diff
@@ -14,7 +14,7 @@ static class AutoStartup
 
         // Don't use Application.ExecutablePath
         // see https://stackoverflow.com/questions/12945805/odd-c-sharp-path-issue
-        
+
         private static string Key = "Shadowsocks_" + Program.ExecutablePath.GetHashCode();
```
studentmain, line 18: string.GetHashCode() are not guranteed same between different versions, it's the possible reason why there's two different entry.

ahmadalli: true, the other issue is that the source string (which we calculate hash of) has been changed from https://github.com/shadowsocks/shadowsocks-windows/commit/ced49e2de12b70908c3ce43a33cb4130fa177879

PR #3048, opened 2020-12-16
remove buggy code

studentmain:

- [ ] Compiled the code with Visual Studio

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

---

remove buggy stream cipher and its test


DuckSoft:
LFGTM

shadowsocks-csharp/Encryption/EncryptorFactory.cs
```diff
@@ -26,7 +26,7 @@ static EncryptorFactory()
             {
                 AEADSodiumEncryptorSupportedCiphers.Remove("aes-256-gcm");
             }
-
+#if I_KNOW_STREAM_CIPHER_IS_UNSAFE
```
DuckSoft, line 29: I'd rather put: if i am sb

DuckSoft (approved):

database64128:
All these overused insecure legacy stream ciphers. Let’s drop them once and for all! 🚀

eebssk1:
I'm using v2ray-pluin TLS while setting SS to use rc4-md5. Is it  really necessary to use AEAD on both plugin and SS itself?

studentmain:
@eebssk1 using v2ray (without plugin) is enough

shadowsocks-csharp/Encryption/EncryptorFactory.cs
fernvenue: > 

So maybe you can tell me why we have to use aead cipher when we already have TLS by [v2ray-plugin](https://github.com/shadowsocks/v2ray-plugin). Or shadowsocks-windows is going to drop support for plugins at a later date?

